PR #8, opened 2023-03-31
Fix unaccept pair request rout bug

GALTdea:
When user with accepted pair_requests visits the pair request index page an error occurs due to the "unaccept" link_to route. In this PR, I  fix the route and link_to helper to fix that issue.


GALTdea:
I think the issue is with the nested resource. I'll comment on the files:

app/views/pair_requests/_pair_request.html.erb
GALTdea, line 1: when the user has accepted pair_requests, this line gets printedlink_to "Unaccept", pair_request_acceptance_path(@pair_request), data: { turbo_method: :delete } ... which renders this:  /pair_requests/:pair_request_id/acceptances/:id(.:format)
I don't believe acceptances can have Id's since they're not models. 

The destroy action in the acceptance_controller is updating the pair_request not destroying a an acceptance record. hope that makes sense!

GALTdea:
Ok, I just added a comment to the specific file, my point is that the route is expecting and id for acceptance which doesn't exist because acceptance is not a persisted model. 
```
  resources :pair_requests do
    resources :acceptances, only: [:create, :destroy]
  end
```
the route rendered by this route is this: 
DELETE /pair_requests/:pair_request_id/acceptances/:id(.:format)

dpaola2 (approved):

toyhammered:
Gonna close this because not relevant anymore.

PR #9, opened 2023-04-03
Issue#2 pair request feedbacck

GALTdea:
As Nathan and I worked on this issue, we ran into the dilemma of whether or not we should add a new model for the rating/feedback. While the pair_request model can store all the info needed for the feature and make our model design less complex, I'm finding that not having ratings/feedback be their model might bring some extra complexity in different ways. 

Not: This  draft request is to get feedback on the overall direction. The code still need review and refactoring.


dpaola2:
Just read through the code and I don't see anything that looks crazy complex to me. Can you be more specific?

In general it's important to distinguish between *easy* and *simple*:

- *easy*: near, at-hand, subjective
- *simple*: unbraided, objective, can be considered in isolation

And, we can distinguish between inherent complexity and incidental complexity:

- inherent complexity: complexity that is inherent in the problem we are solving
- incidental complexity: complexity that we introduce when we make technical decisions

I would say that the complexity of gathering feedback from two participants in a meeting is inherent complexity and, has low complexity. We need to consider the meeting itself and the participants (and nothing else).

GALTdea:
I appreciate the wisdom in your feedback and the importance of being specific with the definitions. 
I agree this is a low-complexity approach, but I wanted to double-check with you!

toyhammered:
Gonna close this because feature has changed.

PR #11, opened 2023-04-12
Adding test env variables

GALTdea:
There is a bug in the app that is impeding it from running rails test as well as the debug gem. Adding test variables to auth.yml fixes the problem.


toyhammered:
Closing because we are not using auth0 anymore.

PR #19, opened 2023-04-27
Disallow Devise New User Registration

### LINKED ISSUE #15: Update Devise to not allow New Registration
Right now anyone can signup for our app. We want to make this invite only for the time being.

JoshDevHub:
## This PR:
- Remove devise's default route configuration for `registrations`. This default setup includes the `/users/sign_up` route.
- Manually adds the routes for editing and updating a user back in. This is because the step above disables all registration routes, but we still want to allow existing user to edit/update their account details.
- Removes existing links to the (now nonexistant) sign up route.
- Removes view for the sign up form, which is now unnecessary because it can't be visited.

## Issue:
Closes #15


app/views/devise/registrations/new.html.erb
toyhammered, line 1: Can we keep this file around because we are prob gonna want it in the future

toyhammered (changes requested):
looks great other than adding that file back!

app/views/devise/registrations/new.html.erb
JoshDevHub: Added it back. It makes sense that we could eventually need this. I was just assuming the invitation library we plan to use had its own set of views to register from.

toyhammered (approved):

PR #20, opened 2023-04-27
PR1 PairRequest Modification

JoshDevHub:
This is step one for the PairRequest feature.

## This PR
- Adds migration for the `pair_requests` table
- Adds associations between `PairRequest` and `User`
- Adds validations for `PairRequest` fields
- Adds tests for more complex validations on the `PairRequest` model


app/models/pair_request.rb
```diff
@@ -2,17 +2,29 @@
...
 class PairRequest < ApplicationRecord
   belongs_to :author, class_name: "User"
-  belongs_to :acceptor, class_name: "User", optional: true
+  belongs_to :invitee, class_name: "User"
 
-  validates :when, presence: true
+  validates :when,
+            presence: true,
+            inclusion: { in: (Date.today..(Date.today + 1.month)) }
+  validates :duration, presence: true, numericality: { greater_than_or_equal_to: 900.0 }
```
toyhammered, line 21: what is the duration value actually (hours?)

app/models/pair_request.rb
```diff
@@ -2,17 +2,29 @@
...
+  validates :status, presence: true
+
```
toyhammered, line 23: We should prob add a `before_create` hook to set this as pending always.

db/schema.rb
```diff
@@ -10,10 +10,20 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_04_26_011733) do
+ActiveRecord::Schema[7.0].define(version: 2023_04_27_204831) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "pair_requests", force: :cascade do |t|
```
toyhammered, line 17: looking at this, the migration seems off. check the format for adding `foreign_key` to a migration.

app/models/user.rb
```diff
@@ -22,4 +22,7 @@ class User < ApplicationRecord
   # NOTE: we will want to add confirmable later on. Will require sendgrid setup (or w/e client we use.)
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :validatable
+
+  has_many :authored_pair_requests, class_name: "PairRequest", foreign_key: "author_id"
+  has_many :received_pair_requests, class_name: "PairRequest", foreign_key: "invitee_id"
```
toyhammered, line 27: for now let's add `dependent: :destroy`. We might want to allow orphans to retain data but not sure.

spec/models/pair_request_spec.rb
```diff
@@ -2,17 +2,36 @@
...
+#  updated_at :datetime         not null
+#  author_id  :integer          not null
+#  invitee_id :integer          not null
 #
 
 require 'rails_helper'
 
 RSpec.describe PairRequest do
-  
+  it "is valid with a duration longer than 15 minutes" do
```
toyhammered, line 18: tonight i'll hop on and change this a bit. I have formatted a little differently when I do these (and you can let me know if we want to go this route). The tests themselves are good though.

toyhammered (changes requested):
super nice! just a few small things.

app/models/pair_request.rb
GALTdea: The duration is 900 seconds wich is 15 minutes

toyhammered: why are we doing seconds?? This should either be minutes (integer) or hours (float). There isn't a purpose to making a meeting 92 seconds. Personally I think it should be minutes and we store as an integer?

app/models/pair_request.rb
GALTdea: hmm, this is on top of the default value in the migration?

toyhammered: ignore this, I feel kinda old because it used to have issues. Yeah when you do `PairRequest.new` it will set the defaults for you.

JoshDevHub:
Alright I think we have all the changes you want in. We also added the `inverse_of` option to the User's `has_many` associations because a) Rubocop wanted it and b) we noticed in the console that queries around those relationships are faster with the `inverse_of` option present.

app/models/pair_request.rb
```diff
@@ -2,17 +2,41 @@
...
+#
+#  fk_rails_...  (author_id => users.id)
+#  fk_rails_...  (invitee_id => users.id)
 #
 class PairRequest < ApplicationRecord
   belongs_to :author, class_name: 'User'
-  belongs_to :acceptor, class_name: 'User', optional: true
+  belongs_to :invitee, class_name: 'User'
+
+  before_create { pending! }
```
toyhammered, line 28: Also, I don't think this is needed (you can add a quick test for creating a new PairRequest and checking that the status is pending.

JoshDevHub: Without the hook, the status will be set to "pending" if the status is unspecified on creation (ie. `nil`).

With your request above for a `before_create` hook, I figured you wanted it such that even if someone did specify the `status` ie:
```rb
PairRequest.create(
  status: "expired",
  # other attrs
)
```
It would still end up being created with the status == "pending". If that's not what you wanted, I can pull it out.

toyhammered: oooooh. I think it should just use the db level status. There are going to be times when we want to create it with something other than pending. Good insight though!

toyhammered: So, lets remove this `before_create`

toyhammered (approved):

Issue #17, opened 2023-04-27
Introduce "role" to a User


toyhammered:
Allow for a user to have a role. For now the 2 roles we need are
1. member (if you have a better name than this I am all ears)
2. admin
This should be an enum.

I don't think we need to go with something like https://github.com/RolifyCommunity/rolify which is just overkill.

dvsconcept1986:
Would like to work on this.

toyhammered:
sounds good, not sure how to assign you but will do so when I figure out.

PR #21, opened 2023-05-01
add background job to auto-reject past pending.

toyhammered:
## Description
This adds a background job to auto-reject any pair requests that are still pending (past the when date).
Also introduces a simple scheduler to help run these jobs, and all the sidekiq configuration.

## Deployment
Before merging this in, I need to setup heroku with redis so that sidekiq works (separate instance from cache).

### DIFF
--- a/app/models/pair_request.rb
+++ b/app/models/pair_request.rb
@@ -27,8 +27,8 @@ class PairRequest < ApplicationRecord
 
   validates :when,
     presence: true,
-    inclusion: { in: (Date.today..(Date.today + 1.month)) }
-  validates :duration, presence: true, numericality: { greater_than_or_equal_to: 15.minutes }
+    inclusion: { in: (Date.current..(Date.current + 1.month)) }
+  validates :duration, presence: true, numericality: { greater_than_or_equal_to: 5.minutes }
   validates :status, presence: true
 
   enum status: {
--- a/spec/factories/pair_requests.rb
+++ b/spec/factories/pair_requests.rb
@@ -26,6 +26,7 @@ FactoryBot.define do
     association :author, factory: :user
     association :invitee, factory: :user
     duration { 15.minutes }
-    add_attribute(:when) { DateTime.now }
+
+    add_attribute(:when) { 1.day.from_now }
   end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,9 @@ gem 'pg', '~> 1.1'
 gem 'puma', '~> 5.0'
 gem 'rails', '~> 7.0.4', '>= 7.0.4.2'
 gem 'redis', '~> 4.0'
+gem 'rufus-scheduler'
 gem 'sassc-rails'
+gem 'sidekiq'
 gem 'sprockets-rails'
 gem 'stimulus-rails'
 gem 'turbo-rails'
@@ -40,6 +42,7 @@ end
 group :test do
   gem 'capybara'
   gem 'faker' # Fake data
+  gem 'rspec-sidekiq'
   gem 'selenium-webdriver'
   gem 'webdrivers'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -93,6 +93,7 @@ GEM
       xpath (~> 3.2)
     coderay (1.1.3)
     concurrent-ruby (1.2.2)
+    connection_pool (2.4.0)
     crass (1.0.6)
     date (3.3.3)
     debug (1.7.1)
@@ -107,6 +108,8 @@ GEM
       warden (~> 1.2.3)
     diff-lcs (1.5.0)
     erubi (1.12.0)
+    et-orbi (1.2.7)
+      tzinfo
     factory_bot (6.2.1)
       activesupport (>= 5.0.0)
     factory_bot_rails (6.2.0)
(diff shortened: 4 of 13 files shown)


spec/factories/pair_requests.rb
```diff
@@ -26,6 +26,16 @@
     association :author, factory: :user
     association :invitee, factory: :user
     duration { 15.minutes }
+
+    # NOTE: figure out how to override these without using a trait....
```
JoshDevHub, line 30: I assume you tried overriding in the job spec, and it didn't work? That seems weird considering how this works in the pair request model spec [here](https://github.com/agency-of-learning/PairApp/blob/main/spec/models/pair_request_spec.rb#L28).

app/jobs/pair_request/auto_reject_job.rb
GALTdea, line 1: hmm What would you think about not changing the status to rejected at the beginning of the session? I would rather allow for at least a few minutes into the session.  In my proposal, I make the update at the end of the proposed end time

GALTdea: maybe we can add set_pair_request_to_rejected method into a helper so we can use it here too?

spec/factories/pair_requests.rb
GALTdea, line 1: Is this test supposed to be passing? mine fails on online 4. We have a  pair_request validation that includes the range of [now .. 1.month] [https://github.com/agency-of-learning/PairApp/blob/main/app/models/pair_request.rb](url)

GALTdea: here's an alternative to using create to instantiate the expired pair request: 
let!(:expired_pending_pair_request) do
    pair_request = build(:pair_request, :past)
    pair_request.save(validate: false)
    pair_request
  end

GALTdea: we'll also have to manually set the status when to "expired"

toyhammered: mine is failing now also... not sure why it passed to begin with lol.

app/jobs/pair_request/auto_reject_job.rb
toyhammered: As for the timing, I truthfully don't think it matters too much when it becomes rejected. Technically speaking we can allow them to update the PairRequest to accepted if they forgot for whatever reason (which can then trigger the feedback)

spec/factories/pair_requests.rb
toyhammered: updated to start passing

app/jobs/pair_request/auto_reject_job.rb
```diff
@@ -0,0 +1,12 @@
+class PairRequest
+  class AutoRejectJob < ApplicationJob
+    sidekiq_options queue: 'low'
+
+    # rubocop:disable Rails/SkipsModelValidations
+    def perform
+      PairRequest.where(status: :pending).where('pair_requests.when < ?',
```
JoshDevHub, line 7: Very small and maybe not worth changing, but enums come with built-in scopes for each of the values. So `PairRequest.where(status: :pending)` can just be `PairRequest.pending`.

toyhammered: great callout! (totally forgot about those)

GALTdea (approved):

toyhammered:
Small blocker while waiting on roles https://github.com/agency-of-learning/PairApp/issues/17

app/jobs/pair_request/auto_reject_job.rb
```diff
@@ -0,0 +1,12 @@
+class PairRequest
+  class AutoRejectJob < ApplicationJob
+    sidekiq_options queue: 'low'
+
+    # rubocop:disable Rails/SkipsModelValidations
+    def perform
+      PairRequest.pending.where('pair_requests.when < ?',
+        Time.current).update_all(status: :rejected)
```
JoshDevHub, line 8: One thing I was thinking about here: We have an `:expired` status for pair requests. To me it might seem more semantically correct to use it for this situation compared to `:rejected`, which feels like a status better kept to when one of the request's owners actively turns the request down.

app/models/pair_request.rb
```diff
@@ -27,8 +27,8 @@ class PairRequest < ApplicationRecord
 
   validates :when,
     presence: true,
-    inclusion: { in: (Date.today..(Date.today + 1.month)) }
-  validates :duration, presence: true, numericality: { greater_than_or_equal_to: 15.minutes }
+    inclusion: { in: (Date.current..(Date.current + 1.month)) }
+  validates :duration, presence: true, numericality: { greater_than_or_equal_to: 5.minutes }
```
JoshDevHub, line 31: Me and Gustavo did have a question about changing this `duration` validation over to 5 minutes minimum. Is that really enough time to hold a pair meeting?

I'm willing to be convinced on this. Just curious to get your thoughts behind it.

toyhammered: I don't think we need to really limit this. You could theoretically have a 5 min just "introduce yourself" with someone. When we do the front-end side, we might just make it intervals of 15 by default and offer a way for them to customize.

app/jobs/pair_request/auto_reject_job.rb
toyhammered: ahh good call, yes i'll change.

app/models/pair_request.rb
JoshDevHub: Ah okay that makes sense. We do have 15 minute intervals in the form and thought dropping down to 5 would make it annoying to use.

PR #29, opened 2023-05-04
Add UI for Pair Requests

JoshDevHub:
I'm opening this as a draft because it's still a bit of a WIP. I wanted to push the code up so @GALTdea could access it and we can work async some. But also I think  it's probably in a good enough spot where we can start accepting some feedback.

## This PR:
- Adds ability to view, create, accept, and reject Pair Requests
- The accepting and rejecting are handled through dedicated controllers.
- The `index` view looks up the pair requests in the context of the `current_user`. This means a user only sees the pair requests they're involved in. A query method had to be written for the direct associations to `PairRequest` don't capture all of the requests a user is involved in. This method is also tested.
- The form for creating pair requests was reworked a little bit to make the select inputs a bit more user friendly.

## Open Questions
- Do we want to tackle time zone stuff in this PR or offload that to another ticket? It's definitely something we want to get right, but we'll also need to talk about how we want to approach it I think.
- The `Acceptances` and `Rejections` controller both have to update the attributes via `#update_attribute` so that validations are bypassed. This is because if the request goes even slightly out of date, we can no longer do something like `pair_request.accepted!` because the validations on the `when` field will fail. We thought it'd be nice to not be super strict on this and allow users to potentially interact with the requests before they're cleaned up by the job that's coming with [this PR](https://github.com/agency-of-learning/PairApp/pull/21). Not sure about this approach though and welcome thoughts on it.
- Also we thought it'd be nice to have the `Acceptances` and `Rejections` controllers namespaced to `PairRequests`. Definitely welcome feedback on that though.
- There's a few spots where logic has leaked into the view. Glad to refactor but not sure what the best approach would be. You can see an example in line 26 of `pair_requests/_form.html.erb`.
- Similarly there's a new `User` method to generate the select input options on the pair request form. Not sure if this really belongs on the `User` model or not and welcome feedback related to it.


.rubocop.yml
```diff
@@ -235,7 +235,7 @@ Style/RescueModifier:
 Style/SelfAssignment:
   Enabled: true
 Style/ClassAndModuleChildren:
-  Enabled: true
+  Enabled: false
```
toyhammered, line 238: Not a fan of this? This is kind of a strange one because of how things worked in the past but I am fine with either.

JoshDevHub: Yeah, this is just because I prefer the way it looks when defining a nested controller or something
```rb
class NameSpace::SomeController < ApplicationController
end

# vs
class NameSpace
  class SomeController
  end
end
```
But I'm willing to be swayed. I also meant to ask about the I18n cops and if we want to turn those off since I don't think that's going to be a concern for us given the scope of this project.

toyhammered: ahh yeah, I'm fine with either so this change is fine. Also yes, we should turn that cop off.

app/views/pair_requests/_form.html.erb
```diff
@@ -11,14 +11,19 @@
...
+
   <div>
     <%= form.label :when, style: "display: block" %>
     <%= form.datetime_field :when %>
   </div>
 
   <div>
     <%= form.label :duration, style: "display: block" %>
-    <%= form.text_field :duration %>
+    <%= form.select :duration, (15..180).step(15).map { |minutes| ["#{minutes} minutes", minutes * 60] } %>
```
toyhammered, line 26: I think we should just make this either a text_field or like input counter (whatever those number things are)

app/views/pair_requests/_pair_request.html.erb
```diff
@@ -5,19 +5,17 @@
...
-        Not yet accepted
-      <% else %>
-        <%= link_to "Accept this request", pair_request_acceptances_path(pair_request), data: { turbo_method: :post }, class: "btn btn-success" %>
-      <% end %>
+    <strong>Invitee: <%= pair_request.invitee.email %></strong>
+  </p>
+
+  <p>
+    <strong>Status: <%= pair_request.status %></strong>
+    <% if pair_request.invitee == current_user && pair_request.pending? %>
```
toyhammered, line 13: This is fine for now. When I integrate pundit in we will go back and fix these up.

toyhammered:
To your open questions:
1. timezone can be separate PR. I have made an issue for it here [#30]
2. That is fine for now. We can also loosen up on the validation. I do like the nested controllers + namespacing.
3. Left some comments on PR, but truthfully it is fine. Once I integrate Pundit that is the route we will go with the views.

Great job though!

app/models/pair_request.rb
```diff
@@ -27,7 +27,9 @@ class PairRequest < ApplicationRecord
 
   validates :when,
     presence: true,
-    inclusion: { in: (Date.current..(Date.current + 1.month)) }
+    inclusion: { in: (Date.current..(Date.current + 1.month)),
+                 message: 'field must not be in the past' }
+  validates :duration, presence: true, numericality: { greater_than_or_equal_to: 15.minutes }
   validates :duration, presence: true, numericality: { greater_than_or_equal_to: 5.minutes }
```
toyhammered, line 33: make sure to remove this cause you are adding another one.

app/policies/application_policy.rb
```diff
@@ -9,31 +9,31 @@ def initialize(user, record)
   end
 
   def index?
-    false
+    user.admin?
```
toyhammered, line 12: I tend to not include this at the application_policy level because there are times where you don't want even admin to have access (and I much prefer being explicit per policy file). You can 100% alias inside the policy file though.

app/views/pair_requests/_pair_request.html.erb
toyhammered: should be able to handle using pundit now.

app/policies/pair_request_policy.rb
```diff
@@ -0,0 +1,34 @@
+class PairRequestPolicy < ApplicationPolicy
+  alias_method :pair_request, :record
+
+  def index?
```
toyhammered, line 4: left comment related to super already. I am thinking how (because our "user" is named differently per records) and we also have an invitee. What i'd like is to have an an `is_owner?` method to consolidate some of these.

app/models/user.rb
```diff
@@ -35,6 +35,12 @@ class User < ApplicationRecord
     dependent: :destroy,
     inverse_of: 'invitee'
 
+  scope :invitee_select_for, ->(user) { User.excluding(user).pluck(:email, :id) }
+
+  def all_pair_requests
```
toyhammered, line 40: i've been thinking what to rename this and maybe something like `my_pair_requests` is more explicit what it actually is?

app/policies/pair_request_policy.rb
```diff
@@ -0,0 +1,48 @@
...
+    true
+  end
+
+  def accept?
+    pair_request.pending? && user == pair_request.invitee
+  end
+
+  def reject?
+    user_is_owner? && (pair_request.pending? || pair_request.accepted?)
+  end
```
toyhammered, line 22: I feel like because we are going with different controllers for different actions to keep things crud, we should do the same with the policy. The class should be like `PairRequest::AcceptancePolicy`. Thoughts on this?

spec/models/user_spec.rb
```diff
@@ -20,6 +20,40 @@
 require 'rails_helper'
 
 RSpec.describe User do
+  describe '::invitee_select_for' do
```
toyhammered, line 23: when handling testing scopes (and naming the blocks) it should either be something like `context 'scopes'` then just a bunch of `describe '.invitee_select_for'`. You can also just do the describe alone. When a method is like `describe '.invitee_select_for'` the `.` implies it is a class method.

spec/policies/pair_request_policy_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    it 'grants access to the user' do
+      expect(subject).to permit(build(:user), build(:pair_request))
+    end
+  end
+
+  permissions :accept? do
+    context 'when the user is not the invitee' do
+      it 'denies access' do
+        authoring_user = build(:user)
+        expect(subject).not_to permit(authoring_user, build(:pair_request, author: authoring_user))
```
toyhammered, line 35: so for these types of tests generally what I do (and how I was taught) is you have a bunch of lets at the top (for like different types of users to use betweeen all the permissions). You can also include some `pair_requests` or scope those directly inside a permission. Typically in rspec you don't want to be building/creating objects but go with the `let` and `let!` approach (and sometimes `before do...`)

Let me know if that makes sense.

app/policies/pair_request_policy.rb
JoshDevHub: I can probably find a workaround, but Pundit is a bit unfriendly to these more "abstract" CRUD operations. For example, in the view, it *needs* this to happen:
```rb
<% if policy(object).action? %>
  <%= render stuff %>
<% end %>
```
But an `Acceptance` isn't an actual domain model as we've written it. The only concrete object I can pass in is a `pair_request` instance, and getting `policy(pair_request)` to look at a different policy than `PairRequestPolicy` seems difficult and potentially impossible. There are "headless" policies, but then those don't use records at all.

But basically I need a policy that can use a pair request as the record, but isn't `PairRequestPolicy` -- and have that integrate with the controllers/views. I couldn't figure it out. If you have ideas for how to go about it, I'd be glad to hear them out.

spec/policies/pair_request_policy_spec.rb
JoshDevHub: It definitely makes sense, and I could write it that way.

This is one of those things that I regularly receive conflicting advice on lol. Some people openly dislike `let` variables in RSpec (including at least one of the seniors in the agency of learning) while others really like them.

But yeah, I figure the important thing is there's a consistent style in the tests, and if that's how you want it, I'll can definitely code it that way.

app/policies/pair_request_policy.rb
GALTdea: hmm, seems like we could namespace the policy. we'd still be using the pair_request object but atleast we'd have a different file for the acceptances_controller?
let me give it a try

toyhammered: Yes to headless. When I was looking through I also thought this code they gave might work:
```
def create
  @publication = find_publication # assume this method returns any model that behaves like a publication
  # @publication.class => Post
  authorize @publication, policy_class: PublicationPolicy
  @publication.publish!
  redirect_to @publication
end
```
We just tell it the policy class (idk if this would work though).

app/views/pair_requests/_pair_request.html.erb
```diff
@@ -5,19 +5,17 @@
...
-        Not yet accepted
-      <% else %>
-        <%= link_to "Accept this request", pair_request_acceptances_path(pair_request), data: { turbo_method: :post }, class: "btn btn-success" %>
-      <% end %>
+    <strong>Invitee: <%= pair_request.invitee.email %></strong>
+  </p>
+
+  <p>
+    <strong>Status: <%= pair_request.status %></strong>
+    <% if  AcceptancePolicy.new(current_user, pair_request).create? %>
```
JoshDevHub, line 13: This is kind of the only way we could find to trigger a policy in the views. The `policy` helper uses metaprogramming to look up the object's policy by name and doesn't offer a way to overwrite this.

Here's an issue referencing this behavior: https://github.com/varvet/pundit/issues/630

toyhammered: lol.... yeah I disagree with what pundit did. We can see how things are going and potentially define a refinement or something to fix this ourselves.

For now what you have is perfect.

spec/policies/pair_request_policy_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  let(:not_owned_request) { build(:pair_request) }
+  let(:authored_request) { build(:pair_request, author: user) }
+  let(:invited_request) { build(:pair_request, invitee: user) }
+  let(:invited_expired_request) { build(:pair_request, invitee: user, status: :expired) }
+
+  permissions :index?, :show? do
+   
+    it 'denies access if the user is not the author, invitee, or an admin account' do
+      binding.b
```
toyhammered, line 16: make sure to remove this

app/policies/pair_request_policy.rb
toyhammered, line 22: remove these now that we have specific policies for them.

app/policies/acceptance_policy.rb
```diff
@@ -0,0 +1,13 @@
+class AcceptancePolicy < ApplicationPolicy
+  attr_reader :user, :pair_request
+
+  def initialize(user, pair_request)
+    @user = user
+    @pair_request = pair_request
+  end
```
toyhammered, line 7: shouldn't need this, it should just get inherited.

toyhammered (commented):
just some removal of uneeded code, otherwise this looks good to me!

app/policies/acceptance_policy.rb
GALTdea: hmm, should we nest the AcceptancePolicy class under PairRequstPolicy? currently we have it under ApplicationPolicy.

GALTdea: we really don't use any of the PairRequestPolicy methods

toyhammered: under ApplicationPolicy is perfect. This file though should most likely be at `app/policies/pair_request/acceptance_policy.rb`

GALTdea: done!

app/policies/pair_request/acceptance_policy.rb
```diff
@@ -0,0 +1,19 @@
+class PairRequest::AcceptancePolicy < ApplicationPolicy
+  attr_reader :user, :pair_request
+
+  def initialize(user, pair_request)
```
JoshDevHub, line 4: If you look in `ApplicationPolicy`, the initializer and `attr_reader` is already set up, and since this class inherits from `ApplicationPolicy`, this stuff doesn't need to be repeated here.

If you look in `PairRequestPolicy` you'll see a nice pattern to use. We just alias the `record` method to be the name of the object we're working with so it's easier to understand exactly what object is involved.

JoshDevHub (commented):
Just a couple of comments, but yeah to me this is close to being done. I can jump in and make these changes later tonight if you want @GALTdea

app/policies/pair_request/acceptance_policy.rb
```diff
@@ -0,0 +1,19 @@
...
+    @user = user
+    @pair_request = pair_request
+  end
+
+  def create?
+    user_is_owner? && (pair_request.pending?)
```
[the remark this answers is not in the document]
GALTdea: Ha good point! Maybe we can make that a paid feature 😆

app/policies/pair_request/rejection_policy.rb
```diff
@@ -0,0 +1,19 @@
+class PairRequest::RejectionPolicy < ApplicationPolicy
+  attr_reader :user, :pair_request
```
toyhammered, line 2: same thing Josh said for this one also.

toyhammered (commented):
this is looking great!

toyhammered (approved):

PR #32, opened 2023-05-05
Add `Procfile.dev` file

### LINKED ISSUE #22: Create dev Procfile
We should add a `Procfile.dev` that will help with setting up our local host. Params we want are:
1. server
2. worker

JoshDevHub:
## This PR:
- Adds a `Procfile.dev` file with instructions for how to run the app's process while in dev environment

## Issue:
Closes #22

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,0 +1,2 @@
+web: unset PORT && bin/rails server
+worker: bundle exec sidekiq -C config/sidekiq.yml
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ To set up this project, you need to have the following software installed on you
 - Ruby 3.1.0
 - Rails 7.0.0 or higher
 - PostgreSQL 9.5 or higher
+- Redis 7.0 or later
 - [Heroku CLI](https://devcenter.heroku.com/articles/heroku-cli) (Follow the link for installation instructions)
 
 ## 12factor Approach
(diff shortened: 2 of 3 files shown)


toyhammered:
> * file with instructions for how to run the app's process while in dev environment

is this missing? (only see 1 file comitted)

JoshDevHub:
> > * file with instructions for how to run the app's process while in dev environment
> 
> is this missing? (only see 1 file comitted)

Oh sorry I mean "instructions" as in a "give the computer instructions for what processes to run" (ie. the content of the Procfile), sorry for the confusion about that.

I did just realize that in order to run the app through `heroku local` (which is currently what the README recommends), we  need to do `heroku local -f Procfile.dev` so the correct Procfile is used. I could shorten that with a `bin/dev` script or something if we want.

toyhammered:
> > > * file with instructions for how to run the app's process while in dev environment
> > 
> > 
> > is this missing? (only see 1 file comitted)
> 
> Oh sorry I mean "instructions" as in a "give the computer instructions for what processes to run" (ie. the content of the Procfile), sorry for the confusion about that.
> 
> I did just realize that in order to run the app through `heroku local` (which is currently what the README recommends), we need to do `heroku local -f Procfile.dev` so the correct Procfile is used. I could shorten that with a `bin/dev` script or something if we want.
> 
> Either way, I'll add some instructions in the README for the new command to use for running the app locally.

adding a `bin/dev` would be great also!

JoshDevHub:
Okay

Added a `bin/dev` script that checks for the Heroku CLI and tells the user to go install it if they don't have it. If they do have the Heroku tooling, it runs `heroku local -f Procfile.dev` and forwards any arguments `bin/dev` received to that command.

Also updated the README to tell devs to use `bin/dev` to run the project. Also felt like it was a good time to go ahead and declare redis as a prerequisite.

toyhammered (approved):

Issue #34, opened 2023-05-07
Frontend Styling Framework


afogel:
The battle between Tailwind or Bootstrap. Who will prevail?! (See design doc)

PR #37, opened 2023-05-10
add initial tailwind config and integrate daisy as component lib

afogel:
Addresses the first part of #34. Subsequent diffs can/will include a refactoring of the existing styling to use DaisyUI components.

## Change in the diff
The diff shows that a new gem `"tailwindcss-rails"` was added to the Gemfile and a new CSS file "application.tailwind.css" was created in the assets/stylesheets directory. 

Additionally, changes were made to the Procfile.dev, config/tailwind.config.js, app/views/layouts/application.html.erb, and app/assets/config/manifest.js files to support the new CSS file. 

These changes involve adding a new process to the development environment to watch for changes to the CSS file, configuring Tailwind CSS, linking the new CSS file in the application layout file, and adding the new CSS file to the asset pipeline manifest.

## How to test:
Checkout branch, `bundle install` and `yarn install`, then run `bin/dev`. It should work.

Open `landing#index.html`. Add a code snippet that utilizes native daisyUI classes, e.g.
```
<div class="btn-primary">
  hello world, I'm a daisy button
</div>
```

When you refresh the page, a purple button should show up.


config/tailwind.config.js
```diff
@@ -0,0 +1,24 @@
...
+  ],
+  theme: {
+    extend: {
+      fontFamily: {
+        sans: ['Inter var', ...defaultTheme.fontFamily.sans],
+      },
+    },
+  },
+  plugins: [
+    require('@tailwindcss/forms'),
```
afogel, line 18: I may be overloading daisyui, but will remove in subsequent diffs

toyhammered:
@afogel got a quick question. Do we need to use any node management. Isn't there a way to purely use rails (and like import maps i think they are called?).

afogel:
> @afogel got a quick question. Do we need to use any node management. Isn't there a way to purely use rails (and like import maps i think they are called?). Not sure if possible because we are including DaisyUi and if so that is fine.

So DaisyUI's installation uses yarn and it seems like using import maps may go against the grain of how libraries like tailwind are best used/may make debugging more challenging later on down the line (see https://github.com/rails/tailwindcss-rails/issues/66).

afogel:
@toyhammered update -- we're currently using import maps. I'm still not loving it, but it's out of scope for this feature request to modify to use a different bundler like esbuild. It'll be fine for now.

toyhammered (approved):

PR #38, opened 2023-05-11
add sendgrid

toyhammered:
## Description
I have added directly to heroku config the proper username + api_key. Need to figure out how we can test this locally/dev. Whoever picks up the devise inviteable it will just be a bit of trial and error (which i'll help with).


JoshDevHub (approved):
LGTM :envelope:

PR #41, opened 2023-05-11
remove some boilerplate files that are not needed

toyhammered:


JoshDevHub (approved):
Looks good to me @toyhammered .

As a mild aside, this stuff will get generated along with any scaffold because of the `jbuilder` gem. Do we want to pull that out? And if we want to keep it around for potential JSON API stuff in the future, we could disable its generator so the scaffold command doesn't invoke it
```rb
# with something like this
config.generators do |g|
  g.jbuilder = false
end
```

toyhammered:
> ```ruby
> config.generators do |g|
>   g.jbuilder = false
> end
> ```

love it! will disable for now. Unsure if we even want to use it for apis.

PR #51, opened 2023-05-13
Build Devise invitable UI

JoshDevHub:
## This PR
- Generates views/mailers for `devise_invitable`
- Styles new forms in accordance with existing login page styles
- Adds an `InvitationPolicy` and uses it to restrict creating invitations to admin users. This policy is also tested.
- Adds letter opener

I ended up doing a couple of things a bit differently than the design doc. One is that actually using/testing this email system is much easier with letter opener, so I just went ahead and added it in. 

The other is that I was originally going to use the builtin controller filter `devise_invitable` has to lock sending invitations to admins only. But that builtin filter is looking at class names to make its determinations, and our app has all users under one class with the `role` field differentiating members and admins. I'm sure I could customize this, but I also wanted to conditionally display a link in the navbar for sending out invitations. At that point, I figured I'd just make a Pundit policy for it and handle the authorization through that.

Maybe one of things I'd specifically like feedback on is:
```rb
class InvitationsController < Devise::InvitationsController
```
Not sure about my controller name here, but a better name didn't immediately occur to me. `ApplicationInvitationsController` maybe?

### DIFF
--- a/spec/policies/invitation_policy_spec.rb
+++ b/spec/policies/invitation_policy_spec.rb
@@ -1,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe InvitationPolicy do
+  subject { described_class }
+
+  let(:member) { build(:user) }
+  let(:admin) { build(:user, :admin) }
+
+  permissions :new?, :create? do
+    it 'denies access to regular member users' do
+      expect(subject).not_to permit(member)
+    end
+
+    it 'grants access to admin users' do
+      expect(subject).to permit(admin)
+    end
+  end
+
+  permissions :edit?, :update? do
+    it 'grants access to member users' do
+      expect(subject).to permit(member)
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -9,9 +9,9 @@ gem 'devise_invitable'
 gem 'image_processing', '~> 1.2'
 gem 'importmap-rails'
 gem 'jbuilder'
+gem 'paper_trail'
 gem 'pg', '~> 1.1'
 gem 'puma', '~> 5.0'
-gem 'paper_trail'
 gem 'rails', '~> 7.0.4', '>= 7.0.4.2'
 gem 'redis', '~> 4.0'
 gem 'rufus-scheduler'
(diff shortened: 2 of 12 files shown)


toyhammered:
> Not sure about my controller name here, but a better name didn't immediately occur to me. `ApplicationInvitationsController` maybe?

That is perfect to me.

spec/policies/invitation_policy_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    it 'denies access to regular member users' do
+      expect(subject).not_to permit(member)
+    end
+
+    it 'grants access to admin users' do
+      expect(subject).to permit(admin)
+    end
+  end
+
+  permissions :edit?, :update? do
```
toyhammered, line 19: What would be a use case for this? Also shouldn't it be scoped to that user directly can only edit/update their own?

toyhammered (commented):
This looks great!

spec/policies/invitation_policy_spec.rb
JoshDevHub: This is kind of due to the way the gem works, I was initially a bit puzzled by it.

If you look at `app/views/devise/invitations/edit.html.erb` you'll see what essentially looks like a sign up page. What happens is a user is created with their email address when the invite is initially sent. When the user clicks on the invite link in their email, they're taken to that "edit" page to enter their password (and potentially `timezone`, etc. in the future) and of course `update` is what happens when they submit.

This flow is already fully locked down by the invitation token. No user can visit the edit page or submit a response to it without a valid invitation token. So I just put `true` there to keep pundit out of the way on that. It's already tightly controlled by the `devise_invitations` gem itself.

toyhammered: ahhh perfect. Thank you for this explanation.

toyhammered (approved):

spec/policies/invitation_policy_spec.rb
GALTdea: Cool, I'll have this flow  in mind when implementing timezone

PR #52, opened 2023-05-13
pin daisyui to importmaps

afogel:
This PR pins daisyui and all dependencies to import maps, which should hopefully enable imports to work (sight unseen).
See whether this works to fix the deploy?

One thing to note: I'm a bit worried about how unwieldy the importmaps might get. All of these changes are the result of running `./bin/importmap pin daisyui`. 

I know that we're not supposed to manage the file manually, instead using the `bin/importmaps` command. That said, in a single file, depending on which JS dependencies we add, it might be painful to work at/understand this file.

Conversely, you might say this is an argument for keeping your dependencies to a minimum, but depending on how interactive we might need the application to be, it might make more sense to add dependencies rather than rolling our own (e.g. flatpickr).

I'm not necessarily advocating anything quite yet, but something to keep an eye on.


toyhammered:
Will be testing this out prob tomorrow.

PR #55, opened 2023-05-15
Adds seed data for user and pair_requet

GALTdea:
This Pr fixes issue 54: “Create Seeds for Users and Pair Requests”. 
A couple of things:
1. I am not great at naming variables so I stack to “admin” “user1”, “user2”, etc. let me know if i should change the naming. 
2. I used fixed data to create users since it’ll make it easy to refer to it in the future. 
3. I added a few pair_requets with some scenarios, but I'm not sure I covered all. Let me know if I’m missing something.


toyhammered (approved):
This is fine for now. Later on we will prob go with just getting an export from staging and using that (so we don't have to keep updating as we add new things and the data is a bit more realistic).

PR #57, opened 2023-05-17
Migrate to esbuild

afogel:
# Context

We decided to use tailwind and daisy for our CSS framework and had been using importmaps to bundle our css and js, but when we tried to deploy heroku, we encountered issues with the `rails assets:precompile` step.

After further investigation, it became evident that this is a dependency issue. Our app currently uses better errors for debugging.

Better errors requires the sassc dependency (https://github.com/BetterErrors/better_errors/issues/516). 
Tailwind uses modern features that are incompatible with sassc and explicitly asks users to remove sassc-rails as a dependency (https://github.com/rails/tailwindcss-rails#conflict-with-sassc-rails). 
Dartsass-rails is a modern replacement for using import maps and sass, but is incompatible with sassc.

# Proposed fix
Keep sassc as a dependency in development for better errors (a la https://github.com/BetterErrors/better_errors/pull/520), but migrate to esbuild for bundling to mitigate subsequent errors.

# What's in this PR?
This PR migrates from using importmaps to esbuild as the default bundling system and updates the README to reflect the fact that getting the app started requires both js and ruby dependencies to be satisfied.


toyhammered (approved):
thank you!

Issue #30, opened 2023-05-04
Introduce timezone to User


toyhammered:
We want to be able to store the timezone on a User.

PR #58, opened 2023-05-18
Pr1 timezone

GALTdea:
Fixes issue #30 this pr will be broken down in 3 parts
1. Pr1: Migration file to add time_zone column to user
2. Pr2 Backfeed user table
3. Pr3 Will add add controllers, and views code to implement timezone
 
This is PR1

### DIFF
--- a/db/migrate/20230516173259_add_column_time_zone_to_users.rb
+++ b/db/migrate/20230516173259_add_column_time_zone_to_users.rb
@@ -1,5 +1,11 @@
 class AddColumnTimeZoneToUsers < ActiveRecord::Migration[7.0]
-  def change
-    add_column :users, :time_zone, :string, default: 'UTC', null: false
-  end
+  def up
+    add_column :users, :time_zone, :string, default: 'UTC'
+    User.update_all(time_zone: 'UTC')
+    change_column :users, :time_zone, :string, null: false
+   end
+   
+   def down
+    remove_column :users, :time_zone
+   end
 end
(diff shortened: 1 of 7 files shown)


db/migrate/20230516173259_add_column_time_zone_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddColumnTimeZoneToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :time_zone, :string, default: 'UTC', null: false
```
toyhammered, line 3: So couple things:
1. is `UTC` correct in what we will be storing. I think they use full names when using the dropdown
2. For this migration it has to be done in steps. Step 1 is create the column with `null: true`. Step 2 is the backfill (in this case you can just do a `User.update_all`. Then you update the column to not allow nulls. You will also need to do `up/down` because the down is just going to be removing the column.

toyhammered (changes requested):

db/migrate/20230516173259_add_column_time_zone_to_users.rb
toyhammered: As a sidenote, for bigger tables you will need to `disable_ttl_transaction!` to make sure the records are not locked (i.e: causea deadlock). Also for bigger tables that require a backfill that you can't do directly in the migration, you will break this up into 2 migrations:
1. creating column with null: true
2. run backfill
3. update column to null: false

GALTdea: I just pushed the changes you asked for. I'll have to do research on disable_ttl_transaction! because i'm not sure about what that is 😅

toyhammered (approved):

PR #59, opened 2023-05-18
Pr2 timezone

GALTdea:
Fixes issue #30 
Pr2:
-Adds params sanitation for the new attribute user time_zone. 
-Adds time_zone_select helper to devise_invitation and devise form

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,14 +2,21 @@ class ApplicationController < ActionController::Base
   include Pundit::Authorization
   include ApplicationHelper
 
+  before_action :configure_permitted_parameters, if: :devise_controller?
+
+
   # PaperTrail helper for getting current_user
   before_action :set_paper_trail_whodunnit
 
-  # around_action :set_time_zone, if: :current_user
+  around_action :set_time_zone, if: :current_user
 
   private
-
   def set_time_zone(&)
     Time.use_zone(current_user.time_zone, &)
+  end
+
+  protected
+  def configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:accept_invitation, keys: [:time_zone])
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -18,5 +18,6 @@ class ApplicationController < ActionController::Base
   protected
   def configure_permitted_parameters
     devise_parameter_sanitizer.permit(:accept_invitation, keys: [:time_zone])
+    devise_parameter_sanitizer.permit(:account_update, keys: [:time_zone])
   end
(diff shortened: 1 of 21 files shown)


app/views/devise/invitations/edit.html.erb
```diff
@@ -15,6 +15,12 @@
     </div>
   <% end %>
 
+  <div class=" mb-8">
+    <%= f.label :time_zone, class: 'label label-text' %>
+    <%= f.time_zone_select(:time_zone, ActiveSupport::TimeZone.country_zones('US'), {}, {class:"select select-bordered w-full max-w-xs"}) %>
```
toyhammered, line 20: We want to allow all country codes. I think there are ways to prioritize some which might be useful?

app/controllers/application_controller.rb
```diff
@@ -2,13 +2,22 @@ class ApplicationController < ActionController::Base
...
   private
   def set_time_zone(&)
     Time.use_zone(current_user.time_zone, &)
   end
+
+  protected
+  def configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:accept_invitation, keys: [:time_zone])
+    devise_parameter_sanitizer.permit(:account_update, keys: [:time_zone])
+  end
```
toyhammered, line 22: very random thing but protected is usually above private for readability reasons.

db/migrate/20230516173259_add_column_time_zone_to_users.rb
```diff
@@ -1,4 +1,7 @@
 class AddColumnTimeZoneToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :time_zone, :string, default: 'UTC', null: false
+  end
```
toyhammered, line 4: this is not right. This should be deleted

db/migrate/20230516173259_add_column_time_zone_to_users.rb
```diff
@@ -8,4 +11,7 @@ def up
    def down
     remove_column :users, :time_zone
    end
+  def change
+    add_column :users, :time_zone, :string, default: 'UTC', null: false
+  end
```
toyhammered, line 16: same here. Looks like your merge had some issues.

toyhammered (changes requested):
This looks great for the most part!

app/views/devise/invitations/edit.html.erb
GALTdea: This setting shows all timezones. the second option `ActiveSupport::TimeZone.country_zones('US')` only prioritizes the US time_zones, it places them on top.

toyhammered (approved):

PR #60, opened 2023-05-19
Update README.md

SyedTathahir:
Update README with current project state and setup instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,27 @@ bin/dev
 
 This will start the development server and background worker. Open your web browser and go to http://localhost:3000 to see the application running.
 
+##Working with Devise Invitable
+If you need to work with the user registration/invitation process, follow these steps:
+
+1. Make sure you are logged out of the application.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,6 +77,27 @@ bin/dev
...
+
+2. Open a Rails console session:
+```bash
+rails console
+```
+3. In the console, run the following command to invite a user:
+```bash 
+User.invite!(email: <some_test_email>)
+```
+This will send an invitation email with an invite link.
```
toyhammered, line 93: Can you include here that for local development we use `letter_opener` and how to access that.

toyhammered (changes requested):
Hey, thanks for contributing this! Would like one small addition otherwise this is great!

README.md
SyedTathahir: I have made changes according to your instructuions  you can review it.

toyhammered (approved):

PR #62, opened 2023-05-22
Create a feedback model with associations

GALTdea:
This pr is the first for the feedback feature

In this pr, we add a Feedback model with polymorphic associations.

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -1,0 +1,32 @@
+# == Schema Information
+#
+# Table name: feedbacks
+#
+#  id                 :bigint           not null, primary key
+#  data               :jsonb
+#  locked_at          :datetime
+#  overall_rating     :integer          default(0), not null
+#  referenceable_type :string           not null
+#  status             :integer          default(0), not null
+#  created_at         :datetime         not null
+#  updated_at         :datetime         not null
+#  author_id          :bigint           not null
+#  receiver_id        :bigint           not null
+#  referenceable_id   :bigint           not null
+#
+# Indexes
+#
+#  index_feedbacks_on_author_id      (author_id)
+#  index_feedbacks_on_receiver_id    (receiver_id)
+#  index_feedbacks_on_referenceable  (referenceable_type,referenceable_id)
+#
+# Foreign Keys
+#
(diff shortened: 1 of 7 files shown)


db/migrate/20230522205955_create_feedbacks.rb
```diff
@@ -0,0 +1,15 @@
+class CreateFeedbacks < ActiveRecord::Migration[7.0]
+  def change
+    create_table :feedbacks do |t|
+      t.references :author, null: false, foreign_key: {to_table: :users}
+      t.references :receiver, null: false, foreign_key: {to_table: :users}
+      t.references :referenceable, polymorphic: true, null: false
+      t.integer :overall_rating, default: 0, null: false
+      t.integer :status, default: 0, null: false
+      t.jsonb :data
```
toyhammered, line 9: should default this to `{}` and null false.

app/models/feedback.rb
```diff
@@ -0,0 +1,32 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_...  (author_id => users.id)
+#  fk_rails_...  (receiver_id => users.id)
+#
+class Feedback < ApplicationRecord
+  belongs_to :author, class_name: "User"
+  belongs_to :receiver, class_name: "User"
+  belongs_to :referenceable, polymorphic: true
```
toyhammered, line 31: should we also include the enum for status in this PR? (0 should be draft)

spec/factories/feedbacks.rb
```diff
@@ -0,0 +1,39 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_...  (author_id => users.id)
+#  fk_rails_...  (receiver_id => users.id)
+#
+FactoryBot.define do
+  factory :feedback do
+    author { nil }
+    receiver { nil }
```
toyhammered, line 31: when you set these up, make sure to point it to a user (should be example in other factory bots).

toyhammered (commented):
This looks amazing! Couple small things to fix but otherwise great job!

JoshDevHub (commented):
Looks good to me. I'll tag @toyhammered just to get his approval for it.

PR #70, opened 2023-05-30
Standup meeting models

toyhammered:
## Description
Added all the tables/models related to creating standup.

I did not use the `create_join_table` because it doesn't create a primary_key and has caused me many issues in the past. This will function the exact same way but we just have more control over it (rather than handing it over to rails).

### DIFF
--- a/app/models/standup_meeting_group_user.rb
+++ b/app/models/standup_meeting_group_user.rb
@@ -1,3 +1,18 @@
+# == Schema Information
+#
+# Table name: standup_meeting_groups_users
+#
+#  id                       :bigint           not null, primary key
+#  created_at               :datetime         not null
+#  updated_at               :datetime         not null
+#  standup_meeting_group_id :bigint           not null
+#  user_id                  :bigint           not null
+#
+# Indexes
+#
+#  index_smg_users_on_smg_id_and_user_id  (standup_meeting_group_id,user_id)
+#  index_smg_users_on_user_id_and_smg_id  (user_id,standup_meeting_group_id)
+#
 class StandupMeetingGroupUser < ApplicationRecord
   self.table_name = 'standup_meeting_groups_users'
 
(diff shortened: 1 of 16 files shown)


db/migrate/20230530025604_create_standup_meetings.rb
```diff
@@ -0,0 +1,17 @@
+class CreateStandupMeetings < ActiveRecord::Migration[7.0]
+  def change
+    create_table :standup_meetings do |t|
+      t.references :standup_meeting_group, null: false, foreign_key: true
+      t.references :user, null: false, foreign_key: true
+
+      t.text :yesterday_work_description, null: false
+      t.text :today_work_description, null: false
+      t.text :blockers_description, null: false
```
JoshDevHub, line 9: Curious if you had any thoughts about how the user will declare that they're not blocked. I guess the intent here might be to have it work like the DailyBot where you have the user put in "no" or "none" or something?

JoshDevHub (approved):
Just a couple of small comments but nothing that's a blocker. Pulled it down and played with the new models in the console and everything seems good to me 🚀

app/models/standup_meeting_group_user.rb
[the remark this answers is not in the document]
toyhammered: just reran `rake db:migrate` and it got it. nice catch!

db/migrate/20230530025604_create_standup_meetings.rb
toyhammered: That was my thought initially but now that I am thinking about it again we should just make this nullable.

toyhammered: Just updated to be nullable. It would also make querying easier because we just check if blockers is not null. We will need to have a tooltip or something to tell the user to leave blank (or maybe just utilize form to show it better)

JoshDevHub: Yeah that's what I was thinking. A nullable field would be pretty convenient to query against for this purpose, so I definitely agree with the change.

PR #71, opened 2023-05-31
update StandupMeeting statuses

toyhammered:
## Description
Just changing the statuses to more closely follow the feedback.


GALTdea (approved):
looks good to me!

PR #77, opened 2023-06-04
UI Build for Feedback Feature

JoshDevHub:
## This PR:
- Adds `FeedbacksController`
- Creates feedback when a user declares a complete Pair Request
- Adds `DATA_OBJECT` for holding a template for the json data that will store the questions and answers in a feedback
- Creates methods to manage the locking of feedback
- Creates Pundit policy for feedback
- Adds edit form for feedback and method to update the object in a way that properly saves to the feedback's jsonb column
- Adds ability to view feedback by visiting `feedbacks#show`

## Items to work on:
This is a draft PR and some things are still a WIP.
- Better integration with Pundit policy and the feedback helper method for generating a feedback link from a pair request. This needs to be accompanied with more test cases for that feedback helper to make sure users don't get an edit link if a feedback is locked.
- The above feedback helper tests can be cleaned up through using factorybot traits on the pair request to create the associated feedback objects.
- Possibly adding places where feedback can be viewed? Currently users are only directly linked to a feedback after they've edited one. Maybe add something to the navbar?


app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  private
+
+  def set_feedback
+    @feedback = Feedback.find(params[:id])
+  end
+
+  def feedback_params
+    params.require(:feedback).permit!
```
toyhammered, line 34: I'm pretty sure you can permit nested params?? Like an array of stuff.

GALTdea:
I think we still need to update the feedback status (from draft to completed) upon submitting feedback

spec/views/feedbacks/edit.html.tailwindcss_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
```
toyhammered, line 1: We can kill all of these tailwindcss files.

app/controllers/feedbacks_controller.rb
JoshDevHub: This was a mistake left in while I was experimenting with things. I added them back in.

app/views/feedbacks/edit.html.erb
```diff
@@ -0,0 +1,33 @@
...
+        <% @feedback.errors.each do |error| %>
+          <li><%= error.full_message %></li>
+        <% end %>
+      </ul>
+    </div>
+  <% end %>
+
+  <div>
+    <%= form.label "overall_rating", class: "label label-text text-lg " %>
+    <%= form.number_field :overall_rating, class: "input input-bordered input-sm" %>
```
toyhammered, line 20: make this required. Also we need to say it is out of `100` (I am also fine if we do out of 10 for simplicity, but when we get the number we multiple by 10 to make it out of 100 when stored.

app/models/pair_request.rb
```diff
@@ -50,4 +49,23 @@ class PairRequest < ApplicationRecord
   def started?
     self.when <= Time.current
   end
+
+  def pending_feedback?(current_user)
+    feedback = references.find_by(author: current_user, status: 'draft')
```
toyhammered, line 54: this can be simplified to just `references.find_by(author: current_user).draft?`

JoshDevHub: Ending on `.draft?` will make it return `true/false`. It could be `references.draft.find_by(author: current_user)` though!

toyhammered: shouldn't this be returning true/false??

GALTdea: Guys, hate to say this, but this method is not being used. This is from before I added the helper 😁

GALTdea: but this method  should return the id of the draft Feedback associated to the user if it exists, nil when it doesn't.

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  def new; end
+
+  def edit
+    authorize @feedback
+  end
+
+  def update
+    authorize @feedback
+
+    if @feedback.update_with_json_answers(params)
```
toyhammered, line 19: this should pass in the status we want to set (or it should just set the status to completed automatically).

spec/helpers/feedbacks_helper_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    let(:pair_request_author) { create(:user) }
+    let(:pair_request_invitee) { create(:user) }
+    let(:random_user) { create(:user) }
+    
+    let(:pair_request) { create(:pair_request, :completed_with_feedback, author_id: pair_request_author.id, invitee_id: pair_request_invitee.id)}
+
+    let(:author_feedback) { pair_request.author_feedback }
+    let(:invitee_feedback) { pair_request.invitee_feedback }
+ 
+    context "when current_user == feedback.author and pair_request status: :draft" do
```
toyhammered, line 13: write these as like `when current_user is the author and the pair_request is a draft` do

spec/helpers/feedbacks_helper_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+   
+      it 'it renders a submit feedback link to the right feedback form ' do
```
toyhammered, line 15: it 'renders a submit feedback link to the correct feedback form'

spec/models/feedback_spec.rb
```diff
@@ -27,6 +27,85 @@
...
+
+  describe '#update_with_json_answers' do
+    let(:params) do
+      ActionController::Parameters.new(
+        feedback: { data:, overall_rating: }
+      )
+    end
+    let(:data) do
+      {
+        'feedback' => {
```
toyhammered, line 43: pref to not use hashrock but symbols: `feedback: [{}]`

app/controllers/feedbacks_controller.rb
GALTdea: hmm, i set the status inside the if block to save if the status is updated. This is in line 20

spec/models/feedback_spec.rb
```diff
@@ -40,7 +40,7 @@
     end
     let(:data) do
       {
-        'feedback' => {
+        feedback: {
           '0' => { 'answer' => 'Answer0' },
```
toyhammered, line 44: Shouldn't the data here be an array? It should try to mimic as closely as possible to real data (could potentially use the const you created?)

GALTdea: yes! just doubled check in the design docs and yes, It should be an array of objects. But does this work better for querying questions?

app/controllers/feedbacks_controller.rb
```diff
@@ -15,9 +15,9 @@ def edit
 
   def update
     authorize @feedback
-
+    @feedback.completed!
```
toyhammered, line 18: @GALTdea this should be inside of the `update_with_json_answers`.

app/models/feedback.rb
```diff
@@ -40,30 +40,30 @@ class Feedback < ApplicationRecord
   # rubocop:disable Layout/LineLength
   DATA_OBJECT = {
     'feedback' => [
-      {
+      strengths: {
```
toyhammered, line 43: This isn't right. it should be like `{ slug: 'strenghts', ...}`

toyhammered (approved):

Issue #73, opened 2023-06-01
Bug: Redirect unauthenticated users to login page


JoshDevHub:
Currently, users aren't redirected to login page if they visit a protected URL without being logged in.

## To recreate:
1. Run the local dev server
2. Be sure to logout if you're logged in
3. Visit `localhost:3000/pair_requests` and you should see the error

## A fix should:
- Redirect any unauthenticated user to the login page if they visit some protected URL (ie one that requires a logged in user). Devise provides a helper called `authenticate_user!` that you can run in a controller filter to implement this.
- While working on this, another change I think would be good is to root unauthenticated users to the login page. So if you visit `localhost:3000/` while not logged in, you should see the login page and not have to click any links/buttons to get there.

Glad to answer any questions!

RocDavenport:
I am interested in this issues.

JoshDevHub:
Apologies on shifting requirements but after looking at some of phase2 for MVP, it looks like parts of the app will eventually be available to public, unauthenticated users -- and we probably don't want to root such users to the login page since they won't have an account.

This just means that we can skip my second bullet point in the issue and just stick to redirecting users off of protected routes if they're not signed in.

Sorry for the change @RocDavenport

elhalvers:
Hey @RocDavenport Let me know if you would like to pair on this issue.

Thanks!

: )

PR #82, opened 2023-06-12
Added authenticate_user! to the pair_requests controller

RocDavenport:
# This PR:

- Fixes the [issue](https://github.com/agency-of-learning/PairApp/issues/73) to Redirect any unauthenticated user to the login page if they visit some protected URL (ie one that requires a logged in user). 
- Added the authenticate_user! helper to the pair_requests_controller as a before action.

# Items to work on:
- Write a systems test for Pair Requests

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -362,6 +362,7 @@ GEM
 PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
+  x86_64-darwin-19
   x86_64-darwin-20
   x86_64-darwin-22
   x86_64-linux
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,8 @@ class ApplicationController < ActionController::Base
   include Pundit::Authorization
   include ApplicationHelper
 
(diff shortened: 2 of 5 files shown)


JoshDevHub (changes requested):
Awesome, congrats on your first PR for the project and big shoutouts to Eric for helping!

So this does solve the visit to something like `/pair_requests`, but there are other protected controllers/routes on the app. And we will end up with more protected routes than unprotected routes. Given that fact, I was thinking we could move this:
```rb
before_action :authenticate_user!
```
To `app/controllers/application_controller.rb`. All controllers in the app inherit from this controller, so by default, every controller action will redirect to the sign in page if the user is not signed in.

The one controller we want publicly open for now would be the `landing_controller` -- we could have the actions in this controller stay public by using a `skip_before_action` so that the `authenticate_user!` method isn't run.

JoshDevHub (approved):
Looks good :rocket:

elhalvers:
Great job @RocDavenport! Excellent comments and suggestion @JoshDevHub! 😃

PR #84, opened 2023-06-13
Pair Request Index Table

JoshDevHub:
## This PR:
- Adds table view for pair requests index list. This "table" is created using tailwind utilities to make it appear table-like but without using actual table elements.
- Adds helper method and date/time formats for formatting a pair requests date to the specifications of in the wireframes
- Adds scopes to pair requests so they can be pulled with the desired ordering (by status, then by date) for the table
- Adds method to look up the partner user for a given user on a pair request


app/helpers/pair_requests_helper.rb
```diff
@@ -1,2 +1,12 @@
 module PairRequestsHelper
+  def format_request_date(pair_request)
```
toyhammered, line 2: fine with this for now but if we start to notice repeated behavior we can refactor then.

toyhammered (approved):

PR #85, opened 2023-06-13
Fix Stimulus Controller Bug - Flash messages not closing

### LINKED ISSUE #78: Bug: Flash messages not closing
There's a bug with flash messages closing. The bug is present for Google Chrome and Safari.

## To Recreate:
1. Start the dev server, visit `localhost:3000`, and login
    - Assuming you've seeded the database properly, you can login with email: `user1@user.com`, password: `password`
2. Try closing the flash message for "Signed in successfully". You should see that it doesn't properly close.

## To Fix:
Make sure flash messages close when you click on their 'X' button. This problem may be related to the `TypeError` that shows up in the browser console, and fixing it may require some adjustments to the way the app's Stimulus controllers are configured.

elhalvers:
## This PR:

- Fixes #78 

## Items to work on:

- Write integration test to test the new stimulus controller functionality.


GALTdea:
feature working and all tests passing locally :) I am getting a weird behaviour though. locally the flash message is closing without a stimulus controller.

GALTdea:
I believe 'alert' tailwindcss-stimulus-components. that's why it works without a defined controller

toyhammered:
> ## Items to work on:
> * Write integration test to test the new stimulus controller functionality.

This is great! Is there a good way to test stimulus controller functionality???

elhalvers:
Hi @GALTdea Yes! I think I just needed to move the code from the 'index' controller to the 'application' controller. Importing and registering the 'tailwindcss-stimulus-components" does the js code work. Without that, we would have had to do it manually by adding the 'alert' controller and defining the '#close' method as I did.

elhalvers:
@toyhammered That's what I have to figure out next! : )

elhalvers:
Just pushed a new commit deleting the unnecessary 'alert' controller.

GALTdea:
@elhalvers Awesome! did you create a test for it? I'll pull and test it locally! Thanks for working on this. going over your code helped me figure out what was wrong with mine

elhalvers:
Hi Gustavo,Cool! Love to hear that!I am researching the testing right now. So far it seems like the way to test is using the jest js testing framework. I’m playing with it and will let you know how it goes.Thanks!: )EricOn Jun 14, 2023, at 12:14, Gustavo Valenzuela ***@***.***> wrote:﻿
@elhalvers Awesome! did you create a test for it? I'll pull and test it locally! Thanks for working on this. going over your code helped me figure out what was wrong with mine

GALTdea:
@elhalvers hmm, not sure you need to test js since we're using an existing library. I would just test that the flash message is being displayed, you could write a request test for that.

elhalvers:
Hi @GALTdea, Thank you for the feedback!

I think testing that the correct flash message is displayed when a user successfully logs in or out might be a good test to have. I notice that there are not currently any functioning 'requests' tests, so maybe this could be a good place to start!

I have not had the chance to read all of the documentation for this project. Has there been any discussions or a design doc around a testing philosophy for the app? Model and System tests, what level of coverage, etc...?

It might make sense to make the 'flash display' test a new 'Let's work on system tests' issue as it does not specifically relate to the changes made in this PR. What do you think?

Maybe later down the line if we continue to use more Stimulus in the front-end it might make sense to include some more js testing to test this functionality.

Thank you Gustavo! Please let me know what you think!

: )

GALTdea:
Another PR for the test would be ok, imo. but if you wanted to add one, I just worked on a request test for casa. Here's the link in case you want a reference. https://github.com/GALTdea/casa/blob/pr3_column_visibility_persistance/spec/requests/preference_sets_spec.rb

toyhammered:
Totally fine with just some manual QA for now if testing is a pain. We can figure it out later.

toyhammered (approved):

PR #90, opened 2023-06-16
Pair request index actions

### LINKED ISSUE #87: Pair Request Index View: Add buttons to 'actions' column
In the current app, the pair requests index table just has some "lorem ipsum" text in the "actions" column. We need to replace this with buttons that perform actions associated with the pair request (ie. Accept, Reject, Review Feedback, etc.)

[See the wireframes here](https://balsamiq.cloud/se3m7v2/puupn9o/r7BA4) (ask for access in the #pairapp discord channel if you don't have it)

Design doc information can be added to the [Frontend - Pair Request doc](https://www.notion.so/toyhammered/Frontend-Pair-Request-a13f23eea820418ba593c4ad1ed90d57)

PR should include:
- The proper buttons/links rendered for each pair request.
- Attached screenshots for demonstrating how it looks.

JoshDevHub:
## This PR:
- Adds and configures view_component (including test setup)
- Adds component for rendering a pair request's status button 
- Refactors how table row links work to support links inside of table cells. Add table cell component to help with this.
- Adds component for rendering an edit feedback button
- Tests new components

## Screenshots
Desktop:
image

Mobile:
image

## Issue
Closes #87


app/controllers/pair_requests/acceptances_controller.rb
```diff
@@ -1,10 +1,12 @@
 class PairRequests::AcceptancesController < ApplicationController
   def create
+    debugger
```
toyhammered, line 3: remove

toyhammered (approved):
Love it! Is there any way for the to get to the pair request show page somehow? Doesn't need to be part of this PR but that is something we should add (because that is also where the feedback is going to be displayed).

PR #92, opened 2023-06-16
hide action column on mobile to improve readability

afogel:
This is a draft proposal for how we might hide the action column on mobile to improve readability. Not sure if this is a best practice when using view components, but I figured I'd throw up a draft and see what y'all think.

Before:
Screen Shot 2023-06-16 at 3 48 51 PM

After:
Screen Shot 2023-06-16 at 3 47 51 PM

### DIFF
--- a/app/components/table_cell_component.rb
+++ b/app/components/table_cell_component.rb
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
 class TableCellComponent < ViewComponent::Base
-  def initialize(container_link: nil)
+  def initialize(container_link: nil, hidden_on_mobile: false)
     @container_link = container_link
+    @hidden_on_mobile = hidden_on_mobile
   end
 
   private
 
-  attr_reader :container_link
+  attr_reader :container_link, :hidden_on_mobile
 end
--- a/app/components/table_cell_component.rb
+++ b/app/components/table_cell_component.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class TableCellComponent < ViewComponent::Base
-  def initialize(container_link: nil, classes: nil)
+  def initialize(container_link: nil, classes: 'table-cell')
     @container_link = container_link
     @classes = classes
   end
--- a/app/components/table_cell_component.rb
+++ b/app/components/table_cell_component.rb
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 
 class TableCellComponent < ViewComponent::Base
-  def initialize(container_link: nil, classes: 'table-cell')
+  def initialize(container_link: nil, class_names: 'table-cell')
     @container_link = container_link
-    @classes = classes
+    @class_names = class_names
   end
 
   private
 
-  attr_reader :container_link, :classes
+  attr_reader :container_link, :class_names
 end
--- a/app/components/table_cell_component.rb
+++ b/app/components/table_cell_component.rb
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 
 class TableCellComponent < ViewComponent::Base
-  def initialize(container_link: nil, hidden_on_mobile: false)
+  def initialize(container_link: nil, classes: nil)
     @container_link = container_link
-    @hidden_on_mobile = hidden_on_mobile
+    @classes = classes
   end
 
   private
 
-  attr_reader :container_link, :hidden_on_mobile
+  attr_reader :container_link, :classes
 end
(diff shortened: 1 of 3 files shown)


app/views/pair_requests/index.html.erb
```diff
@@ -9,7 +9,7 @@
         <div class="table-cell text-left p-1">Partner</div>
         <div class="table-cell text-left p-1">When</div>
         <div class="table-cell text-left p-1">Status</div>
-        <div class="table-cell text-left p-1">Actions</div>
+        <div class="text-left p-1 hidden md:table-cell">Actions</div>
```
JoshDevHub, line 12: This eliminates the `hidden` display at `md` while the cells break are doing it at `sm`. This creates a funky look between these two breakpoints where the column is rendered but without the top label:
image

app/components/table_cell_component.rb
```diff
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
 class TableCellComponent < ViewComponent::Base
-  def initialize(container_link: nil)
+  def initialize(container_link: nil, hidden_on_mobile: false)
```
JoshDevHub, line 4: I wonder if it might be better to inject the styles (or the ones that we expect to vary) from the outside.

A bit nitpicky but with the way you have it, I'm inclined to think there's testing that needs to happen to capture the conditional logic behavior. But with something like:

```rb
class TableCellComponent < ViewComponent::Base
  def initialize(container_link: nil, classes: "table-cell")
    @container_link = container_link
    @classes = classes
  end
end
```
And then template can have some styles we don't expect to change and interpolate them with `classes` that are passed in. Let me know what you think.

toyhammered: Just to chime in here, flags like these are generally a code smell. I'd almost feel like having a component for mobile vs web might be a good way to go? In my draft pr we could go the common route of something like "TableCellComponent.for(device: ,container_link: nil) and initialize either a mobile/web one? Just throwing out some ideas here though.

app/views/pair_requests/index.html.erb
afogel: oops, that was a typo; will fix, thanks for the keen eye!

app/components/table_cell_component.rb
afogel: > I wonder if it might be better to inject the styles (or the ones that we expect to vary) from the outside.
> 
> A bit nitpicky but with the way you have it, I'm inclined to think there's testing that needs to happen to capture the conditional logic behavior. But with something like:
> 
> ```ruby
> class TableCellComponent < ViewComponent::Base
>   def initialize(container_link: nil, classes: "table-cell")
>     @container_link = container_link
>     @classes = classes
>   end
> end
> ```
> 
> And then template can have some styles we don't expect to change and interpolate them with `classes` that are passed in. Let me know what you think.

I 100% agree with both of you -- I threw it up as a proof of concept, figuring that the flag was a smell. I think the difficult here with @toyhammered 's suggestion is that if the window gets resized, the component won't be responsive if it only gets instantiated for a target device. Josh's suggestion is definitely a more robust/elegant solution, will implement and reupload. Thanks!

app/views/pair_requests/index.html.erb
```diff
@@ -3,13 +3,13 @@
 <h1 class="text-2xl font-bold mb-4 text-center">My Requests</h1>
 
 <div class="p-2">
-  <div class="table w-full border-2 border-black text-xs sm:text-sm md:text-base">
+  <div class="table w-full border-2 border-black text-md md:text-base">
```
afogel, line 6: I updated this to improve readability in general

afogel:
Updated screenshots for the changes

Screen Shot 2023-06-19 at 11 57 53 AM
Screen Shot 2023-06-19 at 11 57 45 AM

app/components/table_cell_component.rb
```diff
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
 class TableCellComponent < ViewComponent::Base
-  def initialize(container_link: nil)
+  def initialize(container_link: nil, classes: 'table-cell')
```
toyhammered, line 4: nit thing, but can we call this `class_names`.

toyhammered (changes requested):
Amazing! Honestly one small nit for the variable name (really against `classes`, causes confusion for what it actually is). Once that is fixed i'll merge this in.

app/components/table_cell_component.rb
afogel: super reasonable, updating now.

toyhammered (approved):

PR #100, opened 2023-06-19
[#86] form on pair request index

### LINKED ISSUE #86: Pair Request Index View: Add form for new Pair Request
Currently, when users want to create a new pair request, they must visit a dedicated page with the form on it. Instead of this, users want to be able to fill out a form for a new pair request straight from the pair request index page.

[Wireframes on balsamiq](https://balsamiq.cloud/se3m7v2/puupn9o/r7BA4) (request access in the #pairapp channel on discord if you don't have it)

A PR should include:
- The ability to fill out a form and create a new pair request from the index view. The form design should match the wireframes on desktop and mobile views.
- Attach screenshots in your PR showing the page at both desktop and mobile dimensions.

afogel:
Closes #86 

In this PR, we are:
- rendering the form on the index page
- adding in form styling
- when the form gets submitted, the response should handle either clearing the form or should display the errors associated with the form
- adding in some client-side validation
- prepending successfully completed pair requests to the table

Known issues:

I think there are issues with how the date/time picker currently handles pointer-down events on mobile. The field also has some weird styling at times. I would suggest kicking into a future Issue/PR.

Screenshots:

Happy state:
Screen Shot 2023-06-19 at 3 28 33 PM
Screen Shot 2023-06-19 at 3 28 22 PM

Sad state:

Screen Shot 2023-06-19 at 3 29 20 PM
Screen Shot 2023-06-19 at 3 29 12 PM

Mobile sad state:
Screen Shot 2023-06-19 at 3 29 01 PM


app/controllers/pair_requests_controller.rb
```diff
@@ -27,8 +23,14 @@ def create
           redirect_to pair_request_url(@pair_request),
             notice: 'Pair request was successfully created.'
         end
+        format.turbo_stream do
+          @new_pair_request = @pair_request
+          @pair_request = PairRequest.new
```
toyhammered, line 28: I think it should be `@new_pair_request = PairRequest.new` and `@pair_request` for what is saved.

app/controllers/pair_requests_controller.rb
```diff
@@ -3,6 +3,7 @@ class PairRequestsController < ApplicationController
 
   # GET /pair_requests or /pair_requests.json
   def index
+    @pair_request = PairRequest.new
```
toyhammered, line 6: mentioned below but change to `@new_pair_request`

app/views/pair_requests/_form.html.erb
```diff
@@ -1,32 +1,40 @@
...
+      <div class="alert alert-error flex flex-col items-start my-4">
+        <%= pluralize(pair_request.errors.count, "error") %> prohibited this pair   request from being saved:
+        <ul>
+          <% pair_request.errors.each do |error| %>
+            <li class='font-bold'>
+              <%= error.full_message %>
+            </li>
+          <% end %>
+        </ul>
+      </div>
```
toyhammered, line 13: can we move this whole part into a generic view component called `FormErrorsComponent`?

app/views/pair_requests/create.turbo_stream.erb
```diff
@@ -0,0 +1,32 @@
+<%= turbo_stream.update 'new_request_form' do %>
+  <%= render 'form', pair_request: @pair_request %>
+<% end %>
+
+
+<% if @new_pair_request %>
```
toyhammered, line 6: We shouldn't need this check?

toyhammered (changes requested):
This is fantastic! Only couple things to change. (if you don't agree with something please feel free to leave a comment).

app/controllers/pair_requests_controller.rb
afogel: yeah, agreed, I thought about that after I went out. Kind of whipped it up, but that's an easy and reasonable switch to make.

app/views/pair_requests/_form.html.erb
afogel: I think honestly this should all be shown in a flash. In the spirit of refactoring, I just moved and slightly improved the messaging, but I'd rather that in the `create.turbo_stream.erb` we render a flash message with either errors or success. I think this should be another ticket unto itself, might be another really good first issue/solo.

app/views/pair_requests/create.turbo_stream.erb
afogel: I wasn't sure about that, but figured better safe than sorry. I'll try triggering an error and see whether the request/response cycle complains. If not, better to remove.

app/controllers/pair_requests_controller.rb
afogel: so I realized that this was actually intentional, though I probably could do a better job with naming conventions. By allowing `@pair_request` to stay the same, it lets us keep support the `@pair_request` errors that get shown without having to worry about two cases. I will update the name, however, to better reflect the status of the persisted pair request.

app/views/pair_requests/create.turbo_stream.erb
afogel: After checking, it seems important to keep here. I think we'll be able to remove once we change the error to be a flash stream.

toyhammered (approved):

app/views/pair_requests/create.turbo_stream.erb
afogel: #102 should address this issue

app/views/pair_requests/_form.html.erb
afogel: #102 should address this issue

app/controllers/pair_requests_controller.rb
afogel: #102 should address this issue

PR #101, opened 2023-06-19
Adding turbo to tabs

GALTdea:
Draft PR: 
In this PR @afogel and I pair on
1.  Adding Turbo functionality to the pair_requests table. 
2. Added a daisyUI tab component to the table.
3. Added 2 scopes new scopes: "upcoming" "past"

Turbo is now working appropriately but I run into a couple of issue and questions. 

1. It seems like the when is a SQL [reserved words](https://docs.actian.com/psql/psqlv13/index.html#page/sqlref/sqlkword.htm).  We need to rename that column to be able to execute the right sql queries. if so I suggest the name "pair_request_time"

3. Do we need to add a test for the new turbo feature? if so what kind of test would be good

4. For past and upcoming pair_requests do we just want to rely on status as a determinant?


app/models/pair_request.rb
```diff
@@ -83,4 +83,10 @@ def partner_for(user)
       author
     end
   end
+
+
+  scope :past, -> { where(status: 4) }
+  scope :upcoming, -> { where(status: 2) }
```
JoshDevHub, line 89: I'm a bit curious why these are focused on the status at all? To my mind, it seems like this would just be a future/past query.

app/views/pair_requests/index.html.erb
```diff
@@ -1,42 +1,46 @@
...
-<div class="p-2">
-  <div class="table w-full border-2 border-black text-xs sm:text-sm md:text-base">
-    <div class="table-header-group">
-      <div class="table-row font-bold bg-neutral-400 divide-x-2 divide-black">
-        <div class="table-cell text-left p-1">Partner</div>
-        <div class="table-cell text-left p-1">When</div>
-        <div class="table-cell text-left p-1">Status</div>
-        <div class="table-cell text-left p-1">Actions</div>
-      </div>
+<%=  turbo_frame_tag "pair_requests" do %>
```
JoshDevHub, line 3: Now that this container is a turbo frame, all actions/nav inside of this container are affected by that. If you try visiting any of the links and buttons on the items in the table, you'll see that things are broken. This is because links and buttons inside of a turbo frame expect to use the turbo format, and none of this stuff can handle that as written.

Because it's a completely different page, the links to the pair request show page can probably just use the `data-turbo-frame="_top"` attribute to trigger a full page load.

With buttons like 'Accept', 'Reject', 'Complete', you have a choice. You could just trigger a normal page load. Or you could use turbo to have it where the old buttons are replaced by whatever new thing should be rendered (ie. after a user completes a request, now it should have an 'edit feedback' link). You're free to think that's out of scope for this PR though. It is important to have all the buttons and links working though, whether you get fancy with it or not.

JoshDevHub (commented):
Tabs look and feel nice. Great stuff :muscle:

app/javascript/controllers/hello_controller.js
```diff
@@ -0,0 +1,7 @@
+import { Controller } from "@hotwired/stimulus"
```
afogel, line 1: this file should be removed.

app/javascript/application.js
```diff
@@ -1 +1,2 @@
 import "./controllers";
+import "@hotwired/turbo-rails";
```
afogel, line 2: If you pull from `main` and resolve the conflicts, I think this will already be included

afogel:
Yeah, like you mentioned, `when` is a reserved word. I agree, pair_request_time seems more appropriate than `when` to me.

I don't think we should be testing the view logic of wrapping the table in the `turbo_frame`, though to @JoshDevHub 's point, we need to be careful about how we handle links inside of the frame.

app/views/pair_requests/index.html.erb
```diff
@@ -27,7 +27,7 @@
         <div class="table-row odd:bg-white even:bg-neutral-100 hover:bg-neutral-200 divide-x-2 divide-black">
 
           <%= render TableCellComponent.new(container_link: pair_request) do %>
-            <%= pair_request.partner_for(current_user).email %>
+            <p>John</p><p>Doe</p>
```
afogel, line 30: probably should not use the John Doe placeholder

app/views/pair_requests/index.html.erb
```diff
@@ -38,8 +38,8 @@
             <%= pair_request.status.capitalize %>
           <% end %>
 
-          <%= render(TableCellComponent.new(class_names: 'hidden sm:table-cell')) do %>
-            <%= render PairRequest::StatusButtonsComponent.new(pair_request:, current_user:, style: :link) %>
```
afogel, line 42: I think we want to keep this too, right?

afogel (changes requested):
I think something might have gotten messed up with the forced push, I'm not seeing a lot of work that I'd expect to see here?

app/views/pair_requests/index.html.erb
toyhammered: use email for now

toyhammered:
git issues on this branch. Reference https://github.com/agency-of-learning/PairApp/pull/121 instead

PR #108, opened 2023-06-20
Lint fixes to have a more strict lint check as part of CI

afogel:
This PR represents the lint fixes associated with making the linter be a bit more strict during workflow actions. This will also give us room to discuss which rules to loosen and which ones to keep strict, based on the existing lint failures.


JoshDevHub:
Some of my opinions:
1. Repetitive descriptions and examples should be disabled in `spec/policies/**/*`. This is kind of just the nature of these tests, and it doesn't understand the distinguishing characteristics of things like this:
```rb
  permissions :create?, :update?, :destroy? do
   # collapsing irrelevant code
   
    it 'grants access if the user is an admin' do
      expect(subject).to permit(admin, standup_meeting_group)
    end
  end
  
  permissions :show? do
    # collapsing irrelevant code
    it 'grants access if the user is an admin' do # flags this as a repeated description with the example above
      expect(subject).to permit(admin, standup_meeting_group)
    end
  end
```
2. Multiple assertions should probably be allowed in specs where it makes sense. Certainly system (which we haven't written yet), but I also like them for jobs and components.

3. Memoized helper thing should probably be disabled. Rubocop (the actual gem) disables this for their own [codebase](https://github.com/rubocop/rubocop/blob/master/.rubocop.yml#L156-L157) lol. But also liberally using `let` is kind of an established pattern in how we write tests.

4. The namespaced policies for the `PairRequest` status stuff should probably be fixed. So they can all go in a `spec/policies/pair_request/` directory rather than being at the top level of the policy specs.

5. For line length, I'm not too sure. I personally like to keep lines under 80 chars for most application code, but you basically have to bend on that for certain things. Maybe this can just be disabled on a case by case basis with `disable` comments? Definitely interested in other thoughts on that.

afogel:
lol, I have more thoughts in response to @JoshDevHub 's opinions

1. for the repetitive descriptions, I think it's a good callout by rubocop; I think it suggests that there is a context block that can be reached for. In the example provided, the context block might be:
```ruby
context 'when the user is an admin' do
  permissions :create?, :update?, :destroy?, :show? do
    it 'grants access if the user is an admin' do
      expect(subject).to permit(admin, standup_meeting_group)
    end
  end
end
```
This helps group and emphasize behavior.
2. Agreed. Multiple assertions should be available in component, job, and system tests for sure.
3. I agree with rubocop on this point, too. While we have been using `let` statements liberally, it can end up decoupling state from behavior. In the context of testing, I think localizing state as much as possible can help improve readability of the test and reasoning about test failures, rather than having to scroll to the top of the context block when things go awry. Also, we're often not at a point where we are really duplicating code before we extract that code into a `let` block, so it kind of defeats the purpose of memoizing.
4. I'm on board with fixing the policy stuff too.
5. I don't feel strongly about 80 chars, 100 chars, or 120 chars. Whatever the team prefers is good with me.

JoshDevHub:
> lol, I have more thoughts in response to @JoshDevHub 's opinions
> 
> 1. for the repetitive descriptions, I think it's a good callout by rubocop; I think it suggests that there is a context block that can be reached for. In the example provided, the context block might be:
> 
> ```ruby
> context 'when the user is an admin' do
>   permissions :create?, :update?, :destroy?, :show? do
>     it 'grants access if the user is an admin' do
>       expect(subject).to permit(admin, standup_meeting_group)
>     end
>   end
> end
> ```
> 
> This helps group and emphasize behavior. 3. I agree with rubocop on this point, too. While we have been using `let` statements liberally, it can end up decoupling state from behavior. In the context of testing, I think localizing state as much as possible can help improve readability of the test and reasoning about test failures, rather than having to scroll to the top of the context block when things go awry. Also, we're often not at a point where we are really duplicating code before we extract that code into a `let` block, so it kind of defeats the purpose of memoizing.

Ah yeah the context block is cool way to approach it. I was thinking of the `permissions` block as a bit like a `describe 'method_name'` block, where you probably wouldn't do that type of thing. But I'm probably on board with approaching policy specs in that style.

For `let`, I'd say a way a lot of the tests are written isn't the way I'd naturally write them. Here is an earlier discussion in a PR review about this: https://github.com/agency-of-learning/PairApp/pull/29#discussion_r1188091375

afogel:
> > lol, I have more thoughts in response to @JoshDevHub 's opinions
> > 
> > 1. for the repetitive descriptions, I think it's a good callout by rubocop; I think it suggests that there is a context block that can be reached for. In the example provided, the context block might be:
> > 
> > ```ruby
> > context 'when the user is an admin' do
> >   permissions :create?, :update?, :destroy?, :show? do
> >     it 'grants access if the user is an admin' do
> >       expect(subject).to permit(admin, standup_meeting_group)
> >     end
> >   end
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This helps group and emphasize behavior.
> > 3. I agree with rubocop on this point, too. While we have been using `let` statements liberally, it can end up decoupling state from behavior. In the context of testing, I think localizing state as much as possible can help improve readability of the test and reasoning about test failures, rather than having to scroll to the top of the context block when things go awry. Also, we're often not at a point where we are really duplicating code before we extract that code into a `let` block, so it kind of defeats the purpose of memoizing.
> 
> Ah yeah the context block is cool way to approach it. I was thinking of the `permissions` block as a bit like a `describe 'method_name'` block, where you probably wouldn't do that type of thing. But I'm probably on board with approaching policy specs in that style if others are.
> 

Yeah, I think you're probably still correct about grouping in the example I cited. If I were being more careful about it, I think I'd do it like this:
```ruby
permissions :create?, :update?, :destroy?, :show? do
  context 'when the user is an admin' do
    it 'grants access' do
      expect(subject).to permit(admin, standup_meeting_group)
    end
  end
end
```
I think this helps emphasize still that there is a particular context that is shared for all the policies. Using the `permissions :create?, :update?, :destroy?, :show? do` is a shorthand for grouping them all in a similiar context, anyway.

However, if there are differences between permissions, then I'd say the follow up could be something like:
```ruby
permissions :create? do
  context 'when the user is not an admin' do
    it 'grants access if they are X,Y,Z' do
      expect(subject).to permit(user, standup_meeting_group)
    end
  end
end
```

> For `let`, I'd say the way a lot of the tests are written isn't the way I'd naturally write them, but it was a style we agreed to use earlier on. Here is an earlier discussion in a PR review about this: [#29 (comment)](https://github.com/agency-of-learning/PairApp/pull/29#discussion_r1188091375)

Got it, thanks! I still think I'd like to push some on this, mainly for the reason that Daniel cited about the large files and knowing where state instantiated by `let`s occur. Not that I think that using `let`s is an antipattern, but I don't think it will hurt to be thoughtful about when to abstract out into a memoized var and when not? Perhaps there might be a middle ground by loosening the number of memoized variables used until rubocop starts complaining, maybe to 10-15?

toyhammered:
Ok going through this i'll give my quick opinions:
1. line length is either 100 or 120 (prob 120)
2. I don't like the memoized helper one too much, i'd disable that for now.
3. let blocks one I am fine with (it makes sense)
4. multiple expectations i'd disable for now also. (might re-enable later on?)

Love what you have done though!

afogel:
>     1. line length is either 100 or 120 (prob 120)
Great. 
>     2. I don't like the memoized helper one too much, i'd disable that for now.
> 
>     3. let blocks one I am fine with (it makes sense)
@toyhammered I'm a bit confused -- I was under the impression that the memoized helper and the let blocks were the same. Can you clarify, please? Thanks!

>     4. multiple expectations i'd disable for now also. (might re-enable later on?)
sounds good, will do.

toyhammered:
> 2. I don't like the memoized helper one too much, i'd disable that for now.
> 
> 3. let blocks one I am fine with (it makes sense)

I totally misread! Yeah they are, I don't like how they separate those as 2 different links :sob:. For now let's not use that rule and we can add later on if needed.

spec/factories_spec.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.describe 'FactoryBot' do
   FactoryBot.factories.each do |factory|
     it "#{factory.name} should pass lint" do
-      FactoryBot::Linter.new([factory]).lint!
+      expect(FactoryBot::Linter.new([factory]).lint!).not_to raise(Error)
```
afogel, line 6: rspec examples need to have an assertion, I figured this would be a good way around this error

app/controllers/pair_requests_controller.rb
```diff
@@ -13,6 +13,7 @@ def show
   end
 
   # POST /pair_requests or /pair_requests.json
+  # rubocop:disable Metrics/AbcSize
```
afogel, line 16: I ignored the Assignment Branch Condition size lint error here -- I think the code here is probably fine, but I don't want to disable in general.

JoshDevHub (approved):
Looks good to me! :rocket:

PR #109, opened 2023-06-20
Add Cancelled State

### LINKED ISSUE #53: Add cancelled state to PairRequest
Right now when the user creates a PairRequest (they are the author) they only have the ability to "reject" it afterwards. We should add another state called "cancelled" that is specific to the author.
Author can cancel a PairRequest
Invitee can reject a PairRequest. Invitee can also accept, then reject.

Make sure to update the view to show this change.

afogel:
Closes #53.

Context:
This PR adds in a canceled state that only pair request authors can use on pending or accepted pair requests. 

To accomplish this, the PR:
- adds an enum status onto pair request
- updates the status priorities
- adds a canceled policy
- updates the status button component
- create a namespaced pair_request cancelations controller
- tests all the features above

In addition, the PR also modifies the rejection policy such that an author can no longer reject a pair request.

### DIFF
--- a/spec/components/pair_request/status_buttons_component_spec.rb
+++ b/spec/components/pair_request/status_buttons_component_spec.rb
@@ -3,35 +3,64 @@
 require 'rails_helper'
 
 RSpec.describe PairRequest::StatusButtonsComponent, type: :component do
-  let!(:current_user) { create(:user) }
-  let!(:pending_request) { create(:pair_request, invitee: current_user) }
-  let!(:completed_request) { create(:pair_request, invitee: current_user, status: :completed) }
-  let!(:request_awaiting_completion) do
-    create(:pair_request, author: current_user, when: 5.minutes.ago, status: :accepted)
-  end
+  let(:invitee) { create(:user) }
+  let(:author) { create(:user) }
 
-  context 'with a pair request that can be accepted or rejected' do
-    it 'renders an Accept button and a Reject button' do
-      render_inline(described_class.new(pair_request: pending_request, current_user:))
+  context 'when the current user is the invitee' do
+    let(:current_user) { invitee }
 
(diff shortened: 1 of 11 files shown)


spec/components/pair_request/status_buttons_component_spec.rb
```diff
@@ -3,35 +3,64 @@
...
+
+  context 'when the current user is the invitee' do
+    let(:current_user) { invitee }
 
-  context 'with a pair request that can be accepted or rejected' do
-    it 'renders an Accept button and a Reject button' do
-      render_inline(described_class.new(pair_request: pending_request, current_user:))
+    context 'with a pair request that can be accepted or rejected' do
+      it 'renders an Accept button and a Reject button' do
+        pending_request = create(:pair_request, invitee: current_user)
```
afogel, line 14: vars locally scoped here on purpose, with an eye towards the rationale provided in my lint suggestions: https://github.com/agency-of-learning/PairApp/pull/108#issuecomment-1599679720

spec/policies/cancellation_policy_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'rails_helper'
+RSpec.describe PairRequest::CancellationPolicy do
+  subject { described_class }
+
+  let(:user) { build(:user) }
+  let(:author) { build(:user) }
+  let(:pending_pair_request) { build(:pair_request, author:) }
```
afogel, line 7: the more I think about this, the more I'm not super keen on how many `let` statements I'd like to leverage here, but am happy to leave be or update

toyhammered: I think this is fine.

toyhammered (approved):
This looks good to me!

PR #110, opened 2023-06-21
Modify policy scopes from admin focused to user focused

### LINKED ISSUE #83: Change Pundit scope for "admin" users.
Right now our Pundit Scopes will check if the user is an admin. If they are, it will display all the records (for an index, PairRequest Feedback, etc..). Thinking through, this is not the desired behavior. All scopes should be scoped to the user.

Later on, we will actually create an admin view which will handle displaying these all. Right now admin/member view is kind of muddled which is bad.

elhalvers:
### This Pull Request:
Fixes #83 - Change pundit scope for Admin users.

I paired with Josh today and it seems that this is the fix that might solve the immediate challenges.

### Items to work on:
We noticed that although there are 'policy spec tests' there are not any specific 'Scope' tests. Can create scope tests if desired.

Please let me know if you want to focus on anything else.

Thank you! : )

### DIFF
--- a/app/policies/feedback_policy.rb
+++ b/app/policies/feedback_policy.rb
@@ -28,11 +28,7 @@ class FeedbackPolicy < ApplicationPolicy
     end
 
     def resolve
-      if user.admin?
-        scope.all
-      else
-        user.my_feedback
-      end
+      user.my_feedback
     end
   end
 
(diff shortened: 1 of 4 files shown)


toyhammered (approved):

toyhammered:
This is great, thank you!

Issue #91, opened 2023-06-16
Show partner's time zone to make clear what time is being set


afogel:
As a user, when I create a new Pair Request, I'm asked to select a time/date. However, it's not clear whether I'm selecting a time/date based on my timezone, or my partner's timezone.

**Proposal:**
After I select my partner, whenever I select a date/time, it should be for my local time. Next to the date/time field, when I select a time, it should display what the selected time is in my partner's timezone.

ShamiTomita:
I would love to work on this! Thank you @afogel !

PR #114, opened 2023-06-21
Adding time zones to pair requests form and show

ShamiTomita:
Solves #91 
so far I have been able to create and connect a stimulus controller. I'm able to use the @rails/request.js package to send a get request to the controller which does receive the request and fetches the invitee. Right now I'm having trouble actually receiving and displaying that record in my turbo stream (or that's what I think is happening).


app/views/pair_requests/_form.html.erb
```diff
@@ -12,20 +12,32 @@
         </ul>
       </div>
     <% end %>
-    <div id="form-container" class='flex flex-col space-y-4 xl:space-y-0 xl:max-w-full xl:inline-flex xl:justify-around xl:space-x-8 xl:flex-row'>
+    <div id="form-container" class='flex flex-col space-y-4 xl:space-y-0 xl:max-w-full xl:inline-flex xl:justify-around xl:space-x-8 xl:flex-row' data-controller="pair-requests-form">
       <div class="flex justify-between space-x-4 align-middle">
         <%= form.label :invitee_id, class: "label label-text" %>
         <%= form.select :invitee_id, 
                         User.invitee_select_for(current_user),
```
toyhammered, line 19: Just a thought (which could be entirely wrong), but would it be possible to include the timezone in the `invitee_select_for` call, and then store that as a custom data-* attribute so we don't need to make a seperate call to grab a specific user? Is this a valid javascript option here? (unsure if selects can have data-* options per choice).

app/javascript/controllers/pair_requests_form_controller.js
```diff
@@ -0,0 +1,16 @@
+import { Controller } from "@hotwired/stimulus"
+import { get } from "@rails/request.js"
+// Connects to data-controller="pair-requests-form"
+export default class extends Controller {
```
afogel, line 4: it may not be possible for this, but the hope for stimulus controllers is that they are both focused/small, as well as general enough to be used in multiple places in the app.

I wonder whether it might make sense to refactor this to be more general? Naming it might be hard (I initially thought of `selectOptionTurboStreamTrigger`, but I don't think that's a particularly great name).

One way I might think of making it more general is if the url and param_name are passed to the controller as a `value` attribute, which means you don't need to hardcode it within the controller itself.

This could simplify the change code to be something like:
```js
  change(event) {
    let selectOption = event.target.selectedOptions[0].value

    get(`${urlValue}?${paramValue}=${selectOption},{
      responseKind: "turbo-stream"
    })
  }
```

What do you think?

ShamiTomita: Thank you for the suggestion @afogel, I would definitely want to refactor my approach to reflect that ideal for stimulus controllers to be focused/reusable and your example will really help! I'm trying out some other approaches rn and once I get my ducks in an order I'll see if I can refactor along those lines.

ShamiTomita:
https://github.com/agency-of-learning/PairApp/assets/70528966/b9e10105-ca7a-4e81-8d45-dfed6f24f018

So I think for the most part this issue is done except for two areas that I need some guidance on. 
1. The whole pair request form changes sizes to fit the content which I'm not sure if that is ideals. 
2. Taking the current_user's time zone (their account is set to UTC but they're based in EST for example) makes setting the `when` a little confusing if your computer is not synced to your pair request account. I'm trying to see if I can update the field to completely match the user's time zone but idk if that's feasible or prudent.

app/javascript/controllers/pair_requests_form_controller.js
```diff
@@ -0,0 +1,62 @@
...
+  constructor(input, time) {
+    this.input = input;
+    this.time = time
+  }
+
+  get conversion() {
+    let match
+    let ary = Intl.supportedValuesOf('timeZone');
+    if (this.input == 'UTC'){
+      match = this.input
```
toyhammered, line 47: is this check needed?? Doesn't the else just handle this?

toyhammered (commented):
I mean this seems fine to me but I want to defer this to @afogel and @JoshDevHub

app/javascript/controllers/pair_requests_form_controller.js
ShamiTomita: 'UTC' seemed to be a weird edge case because it's not in JS's `ary = Intl.supportedValuesOf('timeZone');` and without the check the conversion wouldn't be properly set.

app/views/pair_requests/_form.html.erb
```diff
@@ -1,23 +1,32 @@
...
-    <div id="form-container" class='flex flex-col space-y-4 xl:space-y-0 xl:max-w-full xl:inline-flex xl:justify-around xl:space-x-8 xl:flex-row'>
-      <div class="flex justify-between space-x-4 align-middle">
-        <%= form.label :invitee_id, class: "label label-text" %>
-        <%= form.collection_select :invitee_id, User.excluding(current_user), :id, :full_name, {}, { class: 'select select-primary' } %>
+  <%= form_with(model: pair_request, class: 'p-4 mx-auto border-2 border-primary rounded-md flex flex-col') do |form| %>
+    <div id="form-container" class='flex flex-col space-y-4 xl:space-y-0 xl:max-w-full xl:inline-flex xl:justify-around xl:space-x-4 xl:flex-row object-contain' data-controller="pair-requests-form">  
+      <div class="flex justify-between space-x-4 align-middle items-center ">
+         <%= form.label :invitee_id, class: "label label-text" %>
+         <%= form.select :invitee_id, options_for_select(User.excluding(current_user).map{|invitee| [invitee.full_name, invitee.id, {"data-timezone" => invitee.time_zone}]}), {:include_blank => "Please select"}, { class: 'select select-primary', data: {action: "change->pair-requests-form#click"}} %>
```
JoshDevHub, line 6: I think I removed a method out from under you in a different PR. Maybe we should just add it back? The `select` line has become a bit long and difficult to reason through. Could also kick the options to a block passed to `select`, which might look better, something like:
```erb
<%= form.select :invitee_id, nil, { :include_blank => "Please select" }, { class: "select select-primary", data: { action: "change->pair-requests-form#click"} } do %>
  <% User.excluding(current_user).each do |invitee| %>
    <%= tag.option(invitee.full_name, value: invitee.id, data: { timezone: invitee.time_zone}) %>
  <% end %>
<% end %>

**Review by JoshDevHub — COMMENTED** [review]
I think the styling could look cleaner, but I don't really consider that a blocker for merging this. It'd be nice to have the inputs all stay in line and have some way that the new timezone info fits around that without rearranging the datetime field's alignment with other inputs. Perhaps there's something that can be done with grid? Or maybe these [daisy label techniques](https://daisyui.com/components/input/#with-form-control-and-labels)?

But yeah really just the one other comment below. I think I'll wait for Ariel to have a look before merging. I don't see any significant problems or blockers and the styling can be cleaned up whenever.

**Review by afogel — CHANGES_REQUESTED** [review]
Hey! This is a great next pass, great progress 🔥 working with timezones is a headache, so hopefully this isn't just a PITA, but you also feel like you're learning a lot.

So there are still a handful of bugs that I'm noticing with the implementation -- I think the biggest one is that the time zone conversions aren't always working correctly, even in the video you put up (e.g. when the inviter and invitee's timezones are both UTC, they should be the same, when the inviter's timezone is UTC and invitee is EST, they should be different).

Additionally, there are a few harder to catch bugs that we can account for:
The biggest one is a mismatch when the TZ associated with the user's account is different that their computer's TZ - this can be faked from your computer by setting the user timezone to Jerusalem, for example, and seeing how it handles when you're on EST.

Another one has to do with the use of localstorage. As a general heuristic, I reach for local storage when I need to persist data across multiple sessions. In this case, using it causes some weird behavior on refresh since the browser can maintain a value in the pairrequest input field.  This means that if I've selected a date, when I sfot-refresh, changing the user name does not trigger the invitee time w/ timezone to be shown because localstorage was cleared when the controller was reconnected. 
One reasonable alternative method to caching values could be use the `this` keyword to set and get values in the controller, which will remain persisted until the controller gets disconnected.

Alternatively, you can lean more heavily on storing state in the DOM and using stimulus `values`, which is a [recommended practice when writing stimulus controllers](https://stimulus.hotwired.dev/handbook/managing-state).

Even though there's more work to be done, don't get too discouraged! Keep up the good work 💪 !

File: app/views/pair_requests/_form.html.erb (hunk 4 above)
Comment (ShamiTomita) [reply app/views/pair_requests/_form.html.erb:6]: Thank you for the review Josh! I realized earlier that the method was removed so I was trying to work without it but I definitely should've checked in with you about that. I didn't know you could turn the select helper into a block to iterate over the options! It does look a lot better and I'll incorporate that ^_^

File: app/javascript/controllers/pair_requests_form_controller.js [hunk 5]
```diff
@@ -0,0 +1,62 @@
+import { Controller } from "@hotwired/stimulus"
+export default class extends Controller {
+  static targets = ["click", "invitee", "invitee_tz"]
+  connect(){
+    localStorage.clear()
```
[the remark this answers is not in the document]
ShamiTomita: Upon further reflection, this is definitely a heavy handed approach 😅. I need to do a better review of stimulus values so I can utilize that instead of localStorage.

app/views/pair_requests/_form.html.erb
```diff
@@ -1,23 +1,32 @@
...
-        <%= form.label :invitee_id, class: "label label-text" %>
-        <%= form.collection_select :invitee_id, User.excluding(current_user), :id, :full_name, {}, { class: 'select select-primary' } %>
+  <%= form_with(model: pair_request, class: 'p-4 mx-auto border-2 border-primary rounded-md flex flex-col') do |form| %>
+    <div id="form-container" class='flex flex-col space-y-4 xl:space-y-0 xl:max-w-full xl:inline-flex xl:justify-around xl:space-x-4 xl:flex-row object-contain' data-controller="pair-requests-form">  
+      <div class="flex justify-between space-x-4 align-middle items-center ">
+         <%= form.label :invitee_id, class: "label label-text" %>
+         <%= form.select :invitee_id, options_for_select(User.excluding(current_user).map{|invitee| [invitee.full_name, invitee.id, { "data-timezone-identifier" => ActiveSupport::TimeZone.new(invitee.time_zone).tzinfo.identifier, "data-timezone-display-name" => invitee.time_zone, "data-usertz" => current_user.time_zone}]}), {:include_blank => "Please select"}, { class: 'select select-primary', data: {action: "change->pair-requests-form#onClick",  "pair-requests-form-target" => "inviteeInfo"}} %>
```
afogel, line 6: I don't know whether this suggestion is blocking, will defer to @toyhammered . 
This seems like a lot of logic being put in the view, and I think the instantiation of the `ActiveSupport::TimeZone` instance might be the thing pushes it over for me.

I wonder whether this type of logic might be better located in a user decorator, like `Invitee`, and then you can access those invitees using a `@invitees` instance variable?
If you go this route, you'd also need to make sure that a `delegators` folder is slurped up by the application so you'd be able to call it.
e.g.:
```ruby
# app/delegators/invitee.rb
class Invitee < DelegateClass(User)
  def time_zone_identifier
    ActiveSupport::TimeZone.new(time_zone).tzinfo.identifier
  end
end
```
then in the controller, something like
```ruby
# app/controllers/pair_requests_controller.rb
@invitees = User.excluding(current_user).map { |user| Invitee.new(user) }
```

app/views/pair_requests/_form.html.erb
```diff
@@ -1,23 +1,32 @@
...
-      <div class="flex justify-between space-x-4 align-middle">
+      <div class="flex justify-between space-x-4 align-middle items-center">
         <%= form.label :duration, 'Duration (minutes)', class: "label label-text"  %>
         <%= form.number_field :duration, value: 15, in: 15..180, step: 15, required: true, class: 'input input-primary input-md w-36' %>
       </div>
 
-      <div class='ml-auto'>
-        <%= form.submit "Create Pair Request", class: 'btn btn-primary' %>
+      <div class='ml-auto flex items-center'>
+        <%= form.submit "Create Pair Request", class: 'btn btn-primary w' %>
```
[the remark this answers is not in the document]
afogel: see nit

afogel (commented):
Overall, looks great! I did have one comment that I would leave to @toyhammered to decide whether it's blocking re: delegators, and one thing you might be able to do to DRY up your stimulus controller (dayjs extensions).
But overall, it's so much easier to reason about! Really well done!

app/javascript/controllers/pair_requests_form_controller.js
```diff
@@ -0,0 +1,55 @@
...
+    dayjs.extend(timezone)
+    dayjs.extend(customParseFormat)
+
+    const result = dayjs(time).tz(timezoneIdentifier).format("MM/DD/YYYY, hh:mm A")
+    this.inviteeTzTarget.textContent = (`(${timezoneDisplayName})`)
+    this.inviteeTarget.textContent = (`${result}`)
+  }
+
+  setUserSchedule(timezoneIdentifier, time){
+    dayjs.extend(utc)
```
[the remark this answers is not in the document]
ShamiTomita: Ahhh gotcha so right below where I'm importing dayjs would be fine?

app/views/pair_requests/_form.html.erb
ShamiTomita: Oh yeah you're totally right about all the logic here. I still need to implement Josh's suggestion on cleaning it up and I like your suggestion to break it up like that. Thank you Ariel!

ShamiTomita: Hmm, would it be an ok approach to make that delegator method into an instance method for the class? I'm realizing I might need to do the same AS::TZ instantiation for the current user as well.

afogel: I think I was still just wrap the current_user in an `Inviter` delagator where you need it, since the `time_zone_identifier` behavior is context specific to when invitations are necessary, but that's just me. Would be curious to hear what @toyhammered says.

ShamiTomita: Ah ok so in that scenario, in the controller id do something like: `@current_user = Invitee.new(current_user)`

afogel: I guess the question is -- when do you need the current user to have access to the `time_zone_identifier`? how often and for what behavior?

ShamiTomita: That's a very fair question since a lot of the time, we'd only really need to use `current_user.time_zone` when we'd want to display a simple time zone and this identifier is only getting used for the datetime_field conversion.

ShamiTomita:
https://github.com/agency-of-learning/PairApp/assets/70528966/cefce53b-4987-49f5-893e-39c3139c1bbf

app/javascript/controllers/pair_requests_form_controller.js
```diff
@@ -0,0 +1,59 @@
...
+
+export default class extends Controller {
+  
+  static targets = ["invitee", "inviteeTz", "inviterRequestTime", "inviteeInfo",]
+  static values = {
+    input : String
+  }
+
+  onClick(event) {
+    if(this.inviterRequestTimeTarget.value && this.inviteeInfoTarget.value !== ''){
```
afogel, line 18: not blocking (I added an issue for this to incorporate a js linter), but there are style inconsistencies that exist in the file. Typically, we want spaces between an `if` and the `(`.

app/javascript/controllers/pair_requests_form_controller.js
```diff
@@ -0,0 +1,59 @@
...
+      const time = this.inputValue
+      const timeZoneIdentifier = this.inviteeInfoTarget.selectedOptions[0].dataset.timeZoneIdentifier
+      const timeZoneDisplayName = this.inviteeInfoTarget.selectedOptions[0].dataset.timeZoneDisplayName
+
+      this.setInviteeSchedule(timeZoneIdentifier, time, timeZoneDisplayName)
+    }else{
```
afogel, line 24: spaces again (not blocking).

afogel (approved):
lgtm, shipit!

app/javascript/controllers/pair_requests_form_controller.js
ShamiTomita: I see, will keep that in mind for future reference!

app/views/pair_requests/_form.html.erb
```diff
@@ -1,22 +1,35 @@
...
       </div>
 
-      <div class="flex justify-between space-x-4 align-middle">
-        <%= form.label :when, class: "label label-text" %>
-        <%= form.datetime_field :when, required: true, class: "input input-primary input-md" %>
+      <div class="w-full flex flex-col justify-between grow" >
+       <div class=" flex space-x-1 align-middle items-center justify-between">
+          <%= form.label :when, class: "inline-block label label-text pr-2.5" %>
+          <p class="label label-text grow break-words"><%= "(#{current_user.time_zone})" %></p>
+          <%= form.datetime_field :when, required: true, data: {action: "input->pair-requests-form#date", "pair-requests-form-target" => "inviterRequestTime", "pair-requests-form-input-value" => ""}, class: "input input-primary input-md w-max" %>
```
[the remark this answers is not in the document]
ShamiTomita: Yes I can definitely included that in a supplemental Pr. Thank you for your review and suggestions!

Issue #81, opened 2023-06-11
Add user first_name and last_name property to User


GALTdea:
The User model currently lacks a `first_name` and `last_name` property, which limits the amount of information that can be stored and displayed for each user. 

Due to the small table we have, this can be done entirely via 1 migration (single PR). When you create the design doc for this, please do talk about the different tradeoffs of doing this directly via a migration and any other ways this could be handled (if for instance we had a bigger table).

RocDavenport:
I would be interested in taking on this issue.

PR #115, opened 2023-06-22
create a first_name and last_name columns for the user table

RocDavenport:
Issue #81 - 
  - Completed the requirement to add the appropriate columns to the User table. 
  - Addressed back filling via setting a default value.
  - Null has been set to false.


db/migrate/20230621003955_add_names_to_user.rb
```diff
@@ -0,0 +1,6 @@
+class AddNamesToUser < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :first_name, :string, default: 'first_name', null: false
```
toyhammered, line 3: So this is a good start! There are a couple things to note here. This is one potential way to solve this problem but it comes with the downfalls of allowing a default (which we don't want) and locking the db when we run this (which isn't a big issue for our current scale but something to be aware of).

Ima help out a bit here (I realized a bit late that this is a somewhat harder story than I expected). 

What you are gonna want to do locally first is run `rake db:rollback` to revert these changes.

Then update this file to look like this:
```
  def up
    add_column :users, :first_name, :string, null: true
    add_column :users, :last_name, :string, null: true

    User.update_all(first_name: 'first_name', last_name: 'last_name')

    change_column_null :users, :first_name, false
    change_column_null :users, :last_name, false
  end

  def down
    remove_column :users, :first_name, :string
    remove_column :users, :last_name, :string
  end
  ```
  
  What this is doing is:
  1. we add the new columns to the table, but they are nullable
  2. we backfill all the columns to have some "default" data
  3. we make the columns not null because we have filled in all the data
  
  Sidenote: we do this as up/down because we added some additional stuff to be done.

**Review by toyhammered — CHANGES_REQUESTED** [review]
My appologies on this story! I need to figure out a good way to teach everyone how to handle migrations/caveats/backfills.

For this task, I think you should be able to do the second part (part of this PR) which is the update when a user signs up theyu must input their first/last name, and when they go to the my account page, they can edit their first/last name.

Let me know if you have any questions.

File: db/migrate/20230621003955_add_names_to_user.rb (hunk 1 above)
Comment (RocDavenport) [reply db/migrate/20230621003955_add_names_to_user.rb:3]: Thank you for this feedback. Can you tell me exactly what locks the db? I suspect it isn't the adding of the columns, but the back fill via the default value?  I will complete these changes, and it's been really cool to learn more about these differences (and why they are really important for future projects/ uses).

Comment (toyhammered) [reply db/migrate/20230621003955_add_names_to_user.rb:3]: yes, I am pretty sure the `default:` does bad stuff and should NEVER be used as a way to backfill. (maybe things have changed in recent versions, but this is just my general advice).

Comment (RocDavenport) [reply db/migrate/20230621003955_add_names_to_user.rb:3]: Ahh. Thank you for this information.

File: spec/factories/users.rb [hunk 2]
```diff
@@ -5,13 +5,145 @@
...
+#  invited_by_type        :string
+
+#  remember_created_at    :datetime
+#  reset_password_sent_at :datetime
+#  reset_password_token   :string
+#  role                   :integer          default("member"), not null
+#  time_zone              :string           default("UTC"), not null
+#  created_at             :datetime         not null
+#  updated_at             :datetime         not null
+#  invited_by_id          :bigint
```
JoshDevHub, line 102: These annotations are being repeated several times here for some reason, maybe just something funky happening with Annotate. Best to just remove them. Notice the recurring sections beginning with `# Indexes` like on lines 78 and again at 104.

JoshDevHub (changes requested):
Migration looks good to me. I'm sure Daniel will want to get a look at it.

This PR is failing CI due to test issues. It looks like it's probably because there hasn't been any adjustment to `spec/factories/user.rb`. This means the test users created for running the tests are failing the database's not-null constraint.

You'll need to add `first_name` and `last_name` properties in the user factory. Try that and make sure that all tests pass after running `bundle exec rspec`. It's also definitely a good idea to update `db/seeds.rb` so that the database seeds are creating users with names.

If you need/want any  help with this, just let me know on discord

db/migrate/20230621003955_add_names_to_user.rb
```diff
@@ -0,0 +1,16 @@
+class AddNamesToUser < ActiveRecord::Migration[7.0]
+    def up
```
afogel, line 2: nit, but this is indented one tab too many :)

afogel (commented):
overall, looks good! Agreed with @JoshDevHub  about the annotations

db/migrate/20230621003955_add_names_to_user.rb
JoshDevHub: This migration won't work for an important reason @toyhammered 

Devise invitable creates the new user on *invite*, not after the user clicks the link in their email and fills out the form. This means you run straight into a not-null PG error when trying to invite a user just by their email.

To have a not-null constraint with no default, I think we have to allow the admin sending the invite to fill in the user's first and last name while filling out the form to invite them. If that's not desirable, I think we have to make it with a default and just let the user change that when they're signing up.

README.md
```diff
@@ -71,7 +71,7 @@ rails console
 ```
 3. In the console, run the following command to invite a user:
 ```bash 
-User.invite!(email: <some_test_email>)
+User.invite!(email: <some_test_email>, first_name: <user_first_name>, last_name: <user_last_name></user_last_name>)
```
JoshDevHub, line 74: I forgot that we can change this back to just say
```bash
User.invite!(email: <some_test_email>)
```
Because we've added the default name stuff to the `invite!` method.

spec/models/user_spec.rb
```diff
@@ -5,13 +5,145 @@
...
+#  index_users_on_invited_by            (invited_by_type,invited_by_id)
+#  index_users_on_invited_by_id         (invited_by_id)
+#  index_users_on_reset_password_token  (reset_password_token) UNIQUE
+#
+
+#  invitation_accepted_at :datetime
+#  invitation_created_at  :datetime
+#  invitation_limit       :integer
+#  invitation_sent_at     :datetime
+#  invitation_token       :string
```
toyhammered, line 65: I think this was all duplicated.

toyhammered (changes requested):
This is looking great 1 small thing! Also will pull this down today cause wanted to check the account works properly.

JoshDevHub (changes requested):
A few small changes I think should probably be made:
- in `app/models/user.rb`, it's probably a good idea to add a presence validation for the names. This way if the user tries to submit a form with one of the name fields empty, they'll get a good error message and not a 500 page (which is what happens on PG error)

Everything else looks good to me

spec/factories/users.rb
```diff
@@ -31,6 +33,8 @@
 #
 FactoryBot.define do
   factory :user do
+    first_name { Faker::Internet.name }
+    last_name { Faker::Internet.name }
```
JoshDevHub, line 37: So you can check what `Faker` is doing in the console.
```rb
# open a rails console
require 'faker' # be sure to include this line
Faker::Internet.name
#-> "Faker::Internet"
```
This just returns the name of the class, which is standard Ruby behavior
```rb
Array.name
# -> "Array"
```
But we want it to be coming up with a name. Faker has a name module that handles this nicely
```rb
Faker::Name.first_name
# -> "Shirly"
Faker::Name.last_name
# -> "Towne"
```

RocDavenport: This is what I intended on discussing during our call today. I will take a look at these issues.

spec/factories/users.rb
```diff
@@ -31,6 +33,8 @@
 #
 FactoryBot.define do
   factory :user do
+    first_name { Faker::Internet.name.first_name }
```
toyhammered, line 36: This should be the correct format: https://github.com/faker-ruby/faker/blob/main/doc/default/name.md

toyhammered (changes requested):
Everything looks good aside from the failing tests and annotations on the 1 page not being correct.

RocDavenport:
***Please be advised that during development, I noticed functionality that isn't consistent with what my understanding is of what the project calls far. Specifically, the first name and last name fields seem to be hard coded to the default values  where they cannot be changed. I created a user, and then went into my account, and changed the users name from the defaults (first and lsat respectively) to random first and last names. This changes was not saved, and when i logged out and logged back in the default values were still displayed.

JoshDevHub:
> ***Please be advised that during development, I noticed functionality that isn't consistent with what my understanding is of what the project calls far. Specifically, the first name and last name fields seem to be hard coded to the default values where they cannot be changed. I created a user, and then went into my account, and changed the users name from the defaults (first and lsat respectively) to random first and last names. This changes was not saved, and when i logged out and logged back in the default values were still displayed.

Nice catch! @RocDavenport 

If you try to edit your first and last name fields and click the "Update" button, have a look in your server logs and you'll immediately see the problem. There will be something about "Unpermitted Params" -- because you haven't given devise permission to accept these new values for `first_name` and `last_name`.

This is a change that needs to happen in `ApplicationController` for the `configure_permitted_params` method. Currently it's only accepting `time_zone` as an extra parameter for account update and accept invitation. It needs to also accept `first_name` and `last_name` keys.

RocDavenport:
@JoshDevHub i just attempted to change the name while logged in, and indeed received the Unpermitted params error message. 

`Unpermitted parameters: :first_name, :last_name. Context: { controller: Devise::RegistrationsController, action: update, request: #<ActionDispatch::Request:0x000000010f4542b8>, params: {"_method"=>"put", "authenticity_token"=>"[FILTERED]", "user"=>{"first_name"=>"Archester", "last_name"=>"McFlattery III", "email"=>"tester2@testing2.com", "password"=>"[FILTERED]", "password_confirmation"=>"[FILTERED]", "time_zone"=>"London", "current_password"=>"[FILTERED]"}, "commit"=>"Update", "controller"=>"devise/registrations", "action"=>"update"} }`

Made the changes in the controller.

JoshDevHub (approved):
LGTM :rocket:

elhalvers:
Good job @RocDavenport ! 🥳

elhalvers:
I like this migration pattern:

```ruby
Class AddNamesToUser < ActiveRecord::Migration[7.0]
  def up
    add_column :users, :first_name, :string, null: true
    add_column :users, :last_name, :string, null: true

    User.update_all(first_name: 'first_name', last_name: 'last_name')

    change_column_null :users, :first_name, false
    change_column_null :users, :last_name, false
  end
  ```

PR #119, opened 2023-06-23
Display link to create a 'New Standup Group' only for Admins

### LINKED ISSUE #111: Create Standup Group should not be shown to non-admins
Currently, when a non-admin accesses standup_meeting_groups#index, they are able to see a link to standup_meeting_groups#new and even navigate to standup_meeting_groups#new.

Accessing these pages should be restricted to admins.

elhalvers:
### This Pull Request:

- Fixes #111  - Added functionality to the `standup_meeting_groups#index` page to **only** display a link to create a new Standup Meeting Group if the logged in user is an Admin.
- I followed an existing pattern in the app used to hide/display the 'Invite New User' link combining a policy action and a conditional in the view.
- I modified the existing `standup_meeting_group_policy_spec` to include the `:new` action.

### Items to work on:
Can be added to system tests when we implement : )

### DIFF
--- a/app/policies/standup_meeting_group_policy.rb
+++ b/app/policies/standup_meeting_group_policy.rb
@@ -11,10 +11,6 @@ class StandupMeetingGroupPolicy < ApplicationPolicy
     user.admin?
   end
 
-  def new?
-    user.admin?
-  end
-
   def update?
     user.admin?
   end
--- a/app/policies/standup_meeting_group_policy.rb
+++ b/app/policies/standup_meeting_group_policy.rb
@@ -11,7 +11,7 @@ class StandupMeetingGroupPolicy < ApplicationPolicy
     user.admin?
   end
 
-  def edit?
+  def new?
     user.admin?
   end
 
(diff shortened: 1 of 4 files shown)


app/policies/standup_meeting_group_policy.rb
```diff
@@ -11,6 +11,10 @@ def create?
     user.admin?
   end
 
+  def new?
+    user.admin?
+  end
```
JoshDevHub, line 16: We don't need this because of some inheritance stuff happening. If you look in `ApplicationPolicy` (which this class inherits from), you'll see that `new?` looks at what's happening with `create?`, which hopefully makes sense as those actions are closely related.

Given `create?` is defined to look for `user.admin`, you can just remove this and have it inherit the parent's definition for `new?`

JoshDevHub (changes requested):
Looking good! Nice idea to add this check to the tests.

Another thing I think would be good to include with this: although the button is hidden from view if you're not an admin, an ordinary user can still manually visit `/standup_meeting_groups/new`. This is because there isn't controller level authorization in the `new` action for the `StandupMeetingGroupsController`. Probably a good idea to add that in so that only admins can visit that route.

JoshDevHub (approved):
Looks good! Just resolve conflicts and we're good to merge :muscle:

PR #120, opened 2023-06-23
112 edit standup meeting

### LINKED ISSUE #112: [Standup Epic] Create form to create/update standup response
User stories:
As a member of a standup group, I should be able to create a standup response. 

The form should prompt me with the following questions:
1. What did you work on yesterday? (or last working day)
2. What do you plan on working on today?
3. Do you have any blockers?

Additionally, you should have the option to dismiss (`skip`) the standup response for a given date.

See [Standup Meeting (new) ](https://balsamiq.cloud/se3m7v2/puupn9o/r9012) wireframe

afogel:
This PR will close #112 .

There are a few assumptions implicit in this PR. First, a member of a group will always have at least one created standup meeting instance. This will mean that, when a user joins a new group, we should create a standup meeting for that day. All subsequent standups meetings should be created programmatically through a cron job.

As a result, this form also has a min/max date, allowing you to edit standup meetings only for those days that you've got a standup associated with it. This leads to a gaps and islands problem (not all days are consecutive). While I initially thought that might be something that we should kick to another PR, it might make sense for it to live in this PR given the fact that we might not be creating a standup option every day, but rather only on workdays (though, I should note that workweeks are not consistent around the world, e.g. in Israel, I tend to work Sunday thru Thursday).

I also added in flatpickr, since I knew that was on our list of outstanding things to work on (see #1 ). 

**Screenshots:**
Mobile:
Screen Shot 2023-06-23 at 4 54 46 PM

Desktop:
Screen Shot 2023-06-23 at 4 54 15 PM


app/assets/config/manifest.js
```diff
@@ -1,3 +1,2 @@
 //= link_tree ../images
-//= link_directory ../stylesheets .css
```
afogel, line 2: When we use esbuild, it automatically sends the stylesheets to the builds folder. However, to ensure that flatpickr's styles get picked up, we import them in the application.js

This can cause a potential sprockets doublelink error. Therefore, I removed the stylesheets from the manifest.js, which means we will need to explicitly require the css of additional js dependencies if
they have them.

app/components/standup_meeting/action_button_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if policy(:skip).create? %>
+  <% unless standup_meeting.skipped? || standup_meeting.completed? %>
```
toyhammered, line 2: I'm pretty sure ViewCompnent has a `render?` method which can help with this.

app/controllers/standup_meetings/completions_controller.rb
```diff
@@ -0,0 +1,21 @@
+class StandupMeetings::CompletionsController < ApplicationController
+  def create
+    @standup_meeting = StandupMeeting.find(params[:standup_meeting_id])
```
toyhammered, line 3: should be able to just `include` the `standup_meeting_group` to this initial call.

toyhammered (commented):
This is great! Ima test this locally to better understand parts of it!

config/routes.rb
```diff
@@ -32,5 +32,12 @@
     scope module: :standup_meeting_groups do
       resources :joins, only: %i[create destroy]
     end
+    
+    resources :standup_meetings, only: %i[edit update] do
+      scope module: :standup_meetings do
+        resources :skips, only: :create
+        resources :completions, only: :create
+      end
```
toyhammered, line 40: Don't we want this nested under `standup_meeting_groups`?

app/controllers/standup_meetings_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def edit
+    @standup_meeting = StandupMeeting.find(params[:id])
+    # TODO: A naïve date range with a single min and max is probably fine for now,
+    # but there is an islands and gaps problem that should be handled to address
+    # issues when a user joins a group, leaves the group, and rejoins.
+    @date_range = [{
+      from: StandupMeeting.for_member(current_user, @standup_meeting_group).minimum(:meeting_date),
+      to: StandupMeeting.for_member(current_user, @standup_meeting_group).maximum(:meeting_date)
+    }]
```
toyhammered, line 12: I am kind of confused why we need this and what the flow is for editing a standup meeting?

config/routes.rb
toyhammered: thinking about this more we prob don't need to.

app/controllers/standup_meetings_controller.rb
afogel: My recollection during the standup for pair app, you and I discussed that a standup meeting will be created each day for current standup groups through a cron task. I took this to mean that standup meetings should only be available for the days in which you've joined a standup meeting group.

If this is the case, the date range helps us restrict edge cases wherein a user tries to create a standup meeting for a date he wasn't a member of the group. Does that make sense?

app/controllers/standup_meetings_controller.rb
```diff
@@ -0,0 +1,16 @@
+class StandupMeetingsController < ApplicationController
+  before_action :set_standup_meeting_group
+
+  def edit
+    @standup_meeting = StandupMeeting.find(params[:id])
+    authorize @standup_meeting
+  rescue Pundit::NotAuthorizedError
+    redirect_back_or_to(standup_meeting_groups_path)
```
afogel, line 8: new in updated PR.

app/javascript/controllers/application.js
```diff
@@ -13,8 +13,14 @@ application.register('popover', Popover);
 application.register('toggle', Toggle);
 application.register('slideover', Slideover);
 
+
 // Configure Stimulus development experience;
 application.debug = true;
 window.Stimulus   = application;
 
+// import Flatpickr
+import Flatpickr from 'stimulus-flatpickr';
```
afogel, line 22: I'm wondering, since we're planning on using flatpickr soon anyway, when it makes sense to remove since the work is done.

afogel (commented):
@toyhammered ready for rereview, thanks!

app/controllers/standup_meetings_controller.rb
toyhammered: how come we are doing it like this??

app/javascript/controllers/application.js
toyhammered: kind of confused about this comment?

toyhammered (commented):
This is awesome! Just had some questions on comments you left (for clarity on my end).

app/javascript/controllers/application.js
afogel: so I finished implementing the inclusion of flatpickr here, which was slated as a task on the repo (see #1 ). I expect I'll be wanting to reach for it when I build the next set of features, so am wondering whether I need to take it out of this PR here.

app/controllers/standup_meetings_controller.rb
afogel: @toyhammered updated and ready for rereview!

app/controllers/standup_meetings/skips_controller.rb
```diff
@@ -0,0 +1,11 @@
+class StandupMeetings::SkipsController < ApplicationController
+  def create
+    @standup_meeting = StandupMeeting.find(params[:standup_meeting_id])
```
toyhammered, line 3: super small, but this method should have `includes(:standup_meeting_group)` exactly like the completions controller has.

app/controllers/standup_meetings_controller.rb
```diff
@@ -0,0 +1,14 @@
+class StandupMeetingsController < ApplicationController
+  before_action :set_standup_meeting_group
+
+  def edit
+    @standup_meeting = StandupMeeting.find(params[:id])
```
toyhammered, line 5: should just have an `includes` instead of the `before_action`. (and you can set `@standup_meeting_group` in here if needed.

toyhammered (approved):
This is great! 2 small things that need to be fixed, but once those are done feel free to merge this in!

PR #121, opened 2023-06-23
Adding turbo to tabs 2

GALTdea:
Draft PR:

In this PR @afogel and I pair on

Adding Turbo functionality to the pair_requests table.
Added a daisyUI tab component to the table.
Added 2 scopes new scopes: "upcoming" "past"
Turbo is now working appropriately but I run into a couple of issue and questions.

Do we need to add a test for the new turbo feature? if so what kind of test would be good?

"Upcomingt"Pair Request filter
upcominng-filter

"Past" Pair Request filter

past-filter


app/controllers/pair_requests_controller.rb
```diff
@@ -3,8 +3,16 @@ class PairRequestsController < ApplicationController
...
   # GET /pair_requests or /pair_requests.json
   def index
+    if params[:filter] == 'past'
+      @pair_request = PairRequest.new
+      @pair_requests = policy_scope(PairRequest).past.order_by_status.order_by_date
+      render "index", locals: { pair_requests: @pair_requests }
+    else
+      @pair_request = PairRequest.new
+      @pair_requests = policy_scope(PairRequest).upcoming.order_by_status.order_by_date
+    end
```
toyhammered, line 13: As a small exercise try to see if you can refactor all this to be less repetitive.

toyhammered (commented):
This looks great! I do want to pull it down to just test it out a bit and see the feel. Also, for the past/upcoming you should have some type of active link to make it darker.

GALTdea:
Thank you! I realized i was missing the link 'status' after pushing. Should I submit a different pr for that?

toyhammered:
> Thank you! I realized i was missing the link 'status' after pushing. Should I submit a different pr for that?

you can just do as part of this if you want. I'll be merging this in tomorrow sometime most likely (so if not done by then, another PR is fine)

afogel (changes requested):
small changes, but looking good!

app/controllers/standup_meeting_groups_controller.rb
```diff
@@ -12,6 +12,9 @@ def show
 
   def new
     @standup_meeting_group = StandupMeetingGroup.new
+    authorize(@standup_meeting_group)
+  rescue Pundit::NotAuthorizedError
+    redirect_to standup_meeting_groups_url, notice: 'Must be an admin to create a new standup meeting group'
```
toyhammered, line 17: is this needed?? Doesn't this just happen automatically?

app/views/pair_requests/index.html.erb
```diff
@@ -3,29 +3,36 @@
...
-        <% end %>
-        <%= render TableCellComponent.new(class_names: 'hidden sm:table-cell text-center') do %>
-          Actions
-        <% end %>
-      </div>
+<%=  turbo_frame_tag "pair_requests" do %>
+  <div class="p-2">
+    <div class="tabs">
+      <%= link_to 'Past', pair_requests_path(filter: 'past'), data: { turbo_action: 'advance' }, class:"tab tab-lifted #{'tab-active' if params[:filter] == 'past'}"  %>
+      <%= link_to 'Upcoming', pair_requests_path(filter: 'upcoming'), data: { turbo_action: 'advance' }, class:"tab tab-lifted #{'tab-active' if params[:filter] == 'upcoming'}" %>
```
toyhammered, line 10: Upcoming should be above Past.

toyhammered (commented):
This is looking great! Couple small things.

JoshDevHub:
The diff on this PR is super strange to me. For some reason, it's pulling in all sorts of outside changes (like from the standup feature) and adding them to the diff. Hmm

app/controllers/pair_requests_controller.rb
```diff
@@ -4,7 +4,9 @@ class PairRequestsController < ApplicationController
   # GET /pair_requests or /pair_requests.json
   def index
     @pair_request = PairRequest.new
-    @pair_requests = policy_scope(PairRequest).order_by_status.order_by_date
+    filter = params[:filter] == 'past' ? :past : :upcoming
+    @pair_requests = policy_scope(PairRequest).public_send(filter).order_by_status.order_by_date
```
JoshDevHub, line 8: So I think this works great for the upcoming, but I wonder if it may be better for the user if when they are on the 'Past' tab, the records are ordered with the newest ones towards to top of the list? The way it is now, if you're on the past tab, the records furthest into the past are on at the top of the list, but these are the ones the user is less likely to be interacting with.

JoshDevHub (commented):
Just a couple of small comments.

GALTdea:
> The diff on this PR is super strange to me. For some reason, it's pulling in all sorts of outside changes (like from the standup feature) and adding them to the diff. Hmm

I had to run after i submitted the PR but I did feel like there were too many file changes 🤔 i would not merge this until we figure out whats going on

JoshDevHub (commented):
Just one small comment

JoshDevHub:
I think this looks good. I think we could think a bit about how to style the tabs but that's easily done separately, and I don't really see that as a blocker here.

toyhammered (approved):

PR #127, opened 2023-06-26
Edits README.md to include testing and rubocop info

GALTdea:
Just a small PR to add Test and Rubocop information to README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,12 +94,21 @@ git checkout -b your-feature-branch-name
 ```bash
 git add .
 git commit -m "Add your commit message here"
+``` 
+3. Running Test
+```bash
+  bundle exec rspec
 ```
 
-3. Push your changes to origin:
+4. Cleaning up before pushing
+```bash
+  bundle exec rubocop -A
+```
+
+5. Push your changes to origin:
 
 ```bash
 git push origin your-feature-branch-name
 ```
 
-4. Create a new pull request on the original repository and wait for the maintainer to review and merge your changes.
+6. Create a new pull request on the original repository and wait for the maintainer to review and merge your changes.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,12 +94,21 @@ git checkout -b your-feature-branch-name
 ```bash
 git add .
 git commit -m "Add your commit message here"
+``` 
+3. Running Test
```
JoshDevHub, line 98: This would read more naturally with: `3. Running Tests`

I generally don't care to nitpick spelling/English like this but I think it's good to get it correct in a README.

README.md
```diff
@@ -94,12 +94,21 @@ git checkout -b your-feature-branch-name
...
+```bash
+  bundle exec rspec
+```
+
+4. Cleaning up before pushing
+```bash
+  bundle exec rubocop -A
```
JoshDevHub, line 105: I don't think it's a great idea to recommend running rubocop with the `-A` flag. This is because `-A` will run autocorrection over "unsafe cops" -- that is lint offenses where the autocorrection may change how the code functions. I'd probably prefer this to say `bundle exec rubocop -a` and then the developer can manually clean up any remaining offenses.

JoshDevHub (changes requested):
Just a couple of nitpicks. It's a great idea to introduce some instructions around this to the README.

README.md
GALTdea: 100% Good catch!

README.md
GALTdea: Will do! Thanks for the feedback

JoshDevHub (approved):
LGTM :rocket:

PR #130, opened 2023-06-28
Updates to the seed file to include

GALTdea:
This PR  Updates the Seed.rb file to include Pair Requests with "completed" status and with the "when" attribute set in the past. as well as including feedback records for those pair requests.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -17,10 +17,10 @@ begin
       puts "User #{user.email} created"
     end
   end
-
   puts "Users seeded successfully!"
 
   puts "Seeding pair requests..."
+
   pair_request_data = [
     { status: "pending", author: users[1], invitee: users[2], when: 1.day.from_now, duration: 30.minutes },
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -42,9 +41,55 @@
...
+  { status: "accepted", author: users[3], invitee: users[1], when: Time.now - 4.day,  duration: 30.minutes },
+  { status: "accepted", author: users[2], invitee: users[1], when: Time.now - 5.day,  duration: 30.minutes },
+  ]
+  
+  completed_pair_requests = []
+  PairRequest.transaction do
+    completed_pair_request_data.each do |data|
+      pair_request = PairRequest.new(data)
+      pair_request.save(validate: false)
+      # pair_request.create_draft_feedback!
```
JoshDevHub, line 57: Get rid of this comment

JoshDevHub (changes requested):
Just a reminder to get rid of comment. Discussed other changes re: validations in a call.

toyhammered (approved):

PR #131, opened 2023-06-30
113 standup index

afogel:
This is still very much a draft, but this is getting close.

Included currently is the index page, an ability to view either using rows or columns, an ability to navigate to the next day view font-awesome icons, and flatpickr + a stimulus controller to generate a dynamic link builder so that you can jump to a different day.

### DIFF
--- a/app/components/standup_meeting/meeting_update_component.html.erb
+++ b/app/components/standup_meeting/meeting_update_component.html.erb
@@ -1,0 +1,6 @@
+<div class="border p-4 flex flex-col lg:flex-row justify-between mt-4">
+  <span class="flex flex-col justify-center mr-4 font-bold"><%= user.email %></span>
+  <span class="grow">
+    <%= standup_meeting.send(content_type) %>
+  </span>
+</div>
--- a/app/components/standup_meeting/meeting_update_component.rb
+++ b/app/components/standup_meeting/meeting_update_component.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class StandupMeeting::MeetingUpdateComponent < ViewComponent::Base
+  with_collection_parameter :standup_meeting
+
+  def initialize(standup_meeting:, content_type:)
+    @content_type = content_type
+    @standup_meeting = standup_meeting
+    @user = standup_meeting.user
+  end
+
+  private
+
+  attr_reader :content_type, :standup_meeting, :user
+end
--- a/app/helpers/standup_meetings_helper.rb
+++ b/app/helpers/standup_meetings_helper.rb
@@ -1,0 +1,11 @@
+module StandupMeetingsHelper
+  def yesterday(date)
+    return date - 1.day if date.is_a? Date
+    Date.parse(date) - 1.day
+  end
+
+  def tomorrow(date)
+    return date + 1.day if date.is_a? Date
+    Date.parse(date) + 1.day
+  end
+end
--- a/app/views/standup_meetings/index.html.erb
+++ b/app/views/standup_meetings/index.html.erb
@@ -6,33 +6,58 @@
       <li><%= @meeting_date %></li>
     </ul>
   </div>
-  <div class="py-4 flex justify-between ">
-    <h1 class="text-3xl font-bold">Standup - <%= @meeting_date %></h1>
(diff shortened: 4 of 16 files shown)


app/components/standup_meeting/meeting_update_component.html.erb
```diff
@@ -0,0 +1,6 @@
+<div class="border p-4 flex flex-col lg:flex-row justify-between mt-4">
+  <span class="flex flex-col justify-center mr-4 font-bold"><%= user.email %></span>
+  <span class="grow">
+    <%= standup_meeting.send(content_type) %>
```
afogel, line 4: hope this level of metaprogramming is alright.

app/views/standup_meetings/index.html.erb
```diff
@@ -0,0 +1,63 @@
...
+        <i class="fa-solid fa-caret-left h-8"></i> 
+        <span class="md:hidden">
+          Previous
+        </span>
+      <% end %>
+      <%= text_field_tag :date_jumper, @meeting_date, class: 'hidden md:inline border px-4 mx-2 w-48',data: {
+        controller: "flatpickr",
+        flatpickr_date_format: "Y-m-d",
+        flatpickr_max_date: Date.current,
+        action: "change->link-builder#updateDestinationParams",
```
afogel, line 22: on change, this will update the anchor tag right below with new query params

app/components/standup_meeting/meeting_update_component.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class StandupMeeting::MeetingUpdateComponent < ViewComponent::Base
+  with_collection_parameter :standup_meeting
+
+  def initialize(standup_meeting:, content_type:)
+    @content_type = content_type
+    @standup_meeting = standup_meeting
+    @user = standup_meeting.user
```
afogel, line 9: I think because I included the user, n+1 queries should be avoided here? Is this unsafe?

toyhammered: if user was included when making the initial call it should be fine. You can always view the server logs and it'll show you the sql calls you are making (and if you are making multiple of the same)

app/helpers/standup_meetings_helper.rb
```diff
@@ -0,0 +1,11 @@
...
+  def yesterday(date)
+    return date - 1.day if date.is_a? Date
+    Date.parse(date) - 1.day
+  end
+
+  def tomorrow(date)
+    return date + 1.day if date.is_a? Date
+    Date.parse(date) + 1.day
+  end
+end
```
toyhammered, line 11: I feel like a helper isn't what you want here. Also pretty sure active support has like `date.yesterday` or `date.prev_day`. If so, you should use that. 

Gonna think where this can go feels like just a lib though.

app/components/standup_meeting/meeting_update_component.html.erb
toyhammered: My general rule is you should almost never use `send` (exception being inside gems for the most part). You should use `public_send`. For what you are doing i'd honestly move this into a method inside of the component ruby file.

One general alternative to this I like is just having a hash and doing fetch while passing in some param (you can memoize the hash, and it is pretty descriptive in what you are doing).

app/components/standup_meeting/meeting_update_component.rb
afogel: cool, figured as much.

app/helpers/standup_meetings_helper.rb
afogel: yeah -- I just wanted to be careful b/c sometimes it was a date and other times it was a string. I think I handled that in the controller, but...better safe than sorry. Happy to revisit.

app/components/standup_meeting/meeting_update_component.html.erb
afogel: lol. Yeah -- I was on the ropes, which is why I wanted to put this up as a draft. I did have some methods for doing it like a hash, but it introduced a lot more complexity into the component, so I wasn't sure. Anyway, will revisit tomorrow.

PR #136, opened 2023-07-02
Feedback index

### LINKED ISSUE #125: [Feedback Epic] Feedback#index
## Story

As a user, I want the ability to:
- visit a page that has all of my feedback (both given and received) displayed in a list
- click on any individual feedback item to visit its page for more information
- click on a link to edit an individual feedback item if that item is editable by me

## Wireframes
[Balsamiq Link](https://balsamiq.cloud/se3m7v2/puupn9o/r7572)

JoshDevHub:
## Issue
Closes #125 

## Description
- Update feedback's `show?` policy to not authorize draft feedback for receivers
- Small refactor for `Feedback::EditButtonComponent` to allow it to fit more usecases.
- Create `Feedback::ItemComponent` to represent an individual feedback "item" in the list of feedback on a user's index page
- Set up route and controller action for `feedbacks#index`
- Add link to feedback index page in navbar
- Create and style the feedback index page using a tab layout - Stimulus Tab Component used for managing the tab behavior

#### Mobile
Selection_191

#### Desktop
Selection_193

### DIFF
--- a/app/components/feedback/edit_button_component.rb
+++ b/app/components/feedback/edit_button_component.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Feedback::EditButtonComponent < ViewComponent::Base
-  STYLES = {
+  VARIANTS = {
     button: 'btn-primary',
     link: 'btn-link btn-xs sm:btn-sm hover:no-underline'
   }.freeze
@@ -14,11 +14,11 @@ class Feedback::EditButtonComponent < ViewComponent::Base
   end
 
   def call
-    style_archetype = STYLES.fetch(style)
+    variant = VARIANTS.fetch(style)
 
     link_to(
       edit_feedback_path(feedback),
-      class: "btn btn-sm capitalize #{style_archetype} #{classes}"
+      class: "btn btn-sm capitalize #{variant} #{classes}"
     ) do
       content || 'Edit'
     end
--- a/app/components/feedback/edit_button_component.rb
+++ b/app/components/feedback/edit_button_component.rb
@@ -6,11 +6,11 @@ class Feedback::EditButtonComponent < ViewComponent::Base
     link: 'btn-link btn-xs sm:btn-sm hover:no-underline'
   }.freeze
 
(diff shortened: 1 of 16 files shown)


app/components/feedback/edit_button_component.rb
```diff
@@ -1,14 +1,32 @@
...
+  VARIANTS = {
+    button: 'btn-primary',
+    link: 'btn-link btn-xs sm:btn-sm hover:no-underline'
+  }.freeze
+
+  def initialize(feedback:, current_user:, style: :button, classes: '')
     @feedback = feedback
     @current_user = current_user
+    @style = style
+    @classes = classes
```
toyhammered, line 13: make this `class_names` (classes is a confusing word)

app/components/feedback/edit_button_component.rb
```diff
@@ -1,14 +1,32 @@
...
+  end
+
+  def call
+    variant = VARIANTS.fetch(style)
+
+    link_to(
+      edit_feedback_path(feedback),
+      class: "btn btn-sm capitalize #{variant} #{classes}"
+    ) do
+      content || 'Edit'
```
toyhammered, line 23: where does the`content` come from?

toyhammered (changes requested):
This is all great! When displaying the feedback, the order should be desc (latest on top).

app/components/feedback/edit_button_component.rb
JoshDevHub: It's a ViewComponent thing. It comes from a block that's passed to the component. This is a way for the button to have different kinds of text. So if I wanted the button to say "Edit Feedback" instead of just edit ->
```rb
<%= render Feedback::EditButtonComponent(**args) do %>
  Edit Feedback
<% end %>
```
The `content` will be "Edit Feedback" and that's what will get rendered. With no block passed, `content` is `nil`, so I have it default to "Edit".

toyhammered: gotcha! That is really cool!

toyhammered (approved):

PR #139, opened 2023-07-03
Feedback Show

### LINKED ISSUE #126: [Feedback Epic] Feedback#show
Blocked until wireframes are finalized.

## Story

As a user, I'd like to the ability to:
- view a complete instance of feedback (both given and received) that displays the full suite of questions and answers
- click on a button to edit the feedback if I have permissions to do so

JoshDevHub:
## Issue
Closes #126

## Description
- Extracts content on Pair Request show into reusable component
- Uses new component for Pair Request show and Feedback show (if shown feedback is associated with a pair request)

## Screenshots
This is the same content that exists for Pair Request show

#### Mobile
Selection_198

#### Desktop
Selection_197

## Additional Comments
- I think it's probably best to hold off on making this work with User feedback until user feedback is actually implemented as a feature. We could go ahead and do this if we decide that we want to.
- Back linking is also a concern currently. If a user edits/updates feedback, they no longer have a clean way back to the index page from which they came from (browser back is insufficient). This can be solved, but I also think it's something to address with in the PR for feedback edit since the particulars may be dependent on how edit is implemented (ie. turbo stream vs. html response).


toyhammered (approved):
Love it! The one comment I do have is related to who the actual "owner" of this component is. Right now Pair Request owns this, but I am wondering if we want Feedback to be the actual owner? This is just an open question but I am totally good with this PR.

JoshDevHub:
Yeah that's an interesting question. I do think there are some component names that could be redone, but I might stand by this being under the `PairRequest` namespace.

This was originally designed for the pair request show page and shows all the information about the pair request itself + both feedback items associated with the pair request (if they exist). Now those feedback sections are handled by the `Feedback::ContainerComponent` (this name should change I think), but I think this component as a whole is more of a Pair Request thing than a Feedback thing. Willing to be swayed on that though.

PR #145, opened 2023-07-09
add index page for standup meetings

afogel:
Creates general #index page and shell for other features to be added to the StandupMeeting#index page.

Screen Shot 2023-07-09 at 6 42 37 PM

### DIFF
--- a/app/controllers/standup_meetings_controller.rb
+++ b/app/controllers/standup_meetings_controller.rb
@@ -1,7 +1,9 @@
 class StandupMeetingsController < ApplicationController
   def index
     @meeting_date = params[:date].nil? ? Date.current : Date.parse(params[:date])
-    @standup_meeting_group = StandupMeetingGroup.find(params[:standup_meeting_group_id])
+    @standup_meeting_group = StandupMeetingGroup.includes(:users)
+                                                .find(params[:standup_meeting_group_id])
+    authorize @standup_meeting_group, policy_class: StandupMeetingPolicy
     @standup_meetings = @standup_meeting_group.standup_meetings
                                               .includes(:user)
                                               .where(meeting_date: @meeting_date)
--- a/app/controllers/standup_meetings_controller.rb
+++ b/app/controllers/standup_meetings_controller.rb
@@ -1,7 +1,27 @@
 class StandupMeetingsController < ApplicationController
+  # rubocop:disable Metrics/AbcSize
+  def index
+    @meeting_date = params[:date].nil? ? Date.current : Date.parse(params[:date])
+    @standup_meetings = StandupMeeting.includes(:user, :standup_meeting_group)
+                                      .where(
+                                        meeting_date: @meeting_date,
+                                        standup_meeting_group_id: params[:standup_meeting_group_id]
+                                      )
+    @date_options = StandupMeeting.meeting_dates(params[:standup_meeting_group_id])
+    if @standup_meetings.empty?
+      flash[:alert] = 'There are no standups for the date you selected'
+      redirect_to standup_meeting_group_standup_meetings_path(params[:standup_meeting_group_id],
+        date: @date_options.max)
+    else
+      @completed_meetings = @standup_meetings.filter(&:completed?)
+      @current_user_standup_meeting = @standup_meetings.find { |meet| meet.user == current_user }
+      @standup_meeting_group = @standup_meetings.first.standup_meeting_group
+    end
+  end
+  # rubocop:enable Metrics/AbcSize
+
   def edit
-    @standup_meeting = StandupMeeting.includes(:standup_meeting_group)
-                                     .find(params[:id])
+    @standup_meeting = StandupMeeting.includes(:standup_meeting_group).find(params[:id])
     @standup_meeting_group = @standup_meeting.standup_meeting_group
     authorize @standup_meeting
   end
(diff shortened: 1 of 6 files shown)


app/controllers/standup_meetings_controller.rb
```diff
@@ -1,7 +1,25 @@
...
+  # rubocop:disable Metrics/AbcSize
+  def index
+    @meeting_date = params[:date].nil? ? Date.current : Date.parse(params[:date])
+    @standup_meetings = StandupMeeting.includes(:user, :standup_meeting_group)
+                                      .where(
+                                        meeting_date: @meeting_date,
+                                        standup_meeting_group_id: params[:standup_meeting_group_id]
+                                      )
+    @date_options = StandupMeeting.meeting_dates(params[:standup_meeting_group_id])
+    if @standup_meetings.empty?
```
afogel, line 11: gracefully handle the event in which a particular date requested by a user does not have any StandupMeetings that fall on that date.

app/controllers/standup_meetings_controller.rb
```diff
@@ -1,7 +1,23 @@
 class StandupMeetingsController < ApplicationController
+  def index
+    @meeting_date = params[:date].nil? ? Date.current : Date.parse(params[:date])
+    @standup_meetings = StandupMeeting.includes(:user, :standup_meeting_group)
```
toyhammered, line 4: I think this should be changed to something like this:
```
@standup_meeting_group = StandupMeetingGroup.find(standup_meeting_group_id)
@standup_meetings = @standup_meeting_group.standup_meetings.includes(:user).where(meeting_date: @meeting_date)
```
The main reason is we should still go to that day even if there are no standup meetings for it. It should be treated the exact same  and on the view side it will just be empty in that bottom section.

app/controllers/standup_meetings_controller.rb
```diff
@@ -1,7 +1,23 @@
 class StandupMeetingsController < ApplicationController
+  def index
+    @meeting_date = params[:date].nil? ? Date.current : Date.parse(params[:date])
```
toyhammered, line 3: I am debating on whether this should be a required param and the routes pass it in (in like 95% of the cases they will).

app/controllers/standup_meetings_controller.rb
```diff
@@ -1,7 +1,23 @@
...
+    @standup_meetings = StandupMeeting.includes(:user, :standup_meeting_group)
+                                      .where(
+                                        meeting_date: @meeting_date,
+                                        standup_meeting_group_id: params[:standup_meeting_group_id]
+                                      )
+    @date_options = StandupMeeting.meeting_dates(params[:standup_meeting_group_id])
```
toyhammered, line 9: Related to the above, I don't think we even need this if this was only used in the redirect due to there being no meetings on that day.

toyhammered (changes requested):
Let me know if you have any questions. I'd like to keep this still very simple.

app/controllers/standup_meetings_controller.rb
afogel: This is used later for moving between dates in https://github.com/agency-of-learning/PairApp/pull/146 as well. I introduced it here because it is also useful for the redirect guard.

toyhammered: I don't think we need it there either. The logic should just be +1 or -1 to the current date. Let weekend pages show with no standup meetings.

app/controllers/standup_meetings_controller.rb
afogel: I think it probably makes sense to keep, just in case?

app/controllers/standup_meetings_controller.rb
afogel: This makes a ton more sense to me -- kind of incredible that such a small change simplifies things so dramatically.

app/policies/standup_meeting_policy.rb
```diff
@@ -1,8 +1,6 @@
 class StandupMeetingPolicy < ApplicationPolicy
-  alias_method :standup_meeting, :record
-
   def index?
-    user.admin? || matching_user?
+    group_member? || user.admin?
```
toyhammered, line 3: fine with this for now but this is a strange one cause of the mixing/matching of what the "record" actually is. Not sure of a good solution to make this clearer.

app/policies/standup_meeting_policy.rb
```diff
@@ -12,12 +10,10 @@ def update?
...
-    standup_meeting.user_id == user.id
+    record.user_id == user.id
   end
 
-  class Scope < Scope
-    def resolve
-      user.standup_meetings
-    end
+  def group_member?
+    record.users.where(id: user.id).present?
```
toyhammered, line 17: should just do `find_by` because we are looking for a single record. Technically doesn't make much of a difference just more clear.

toyhammered (approved):
Just one small thing to fix, but once that is done feel free to merge in.

app/policies/standup_meeting_policy.rb
JoshDevHub: Since it's a boolean check for one record, I really like to reach for `#exists?` [Rails guide for it](https://guides.rubyonrails.org/active_record_querying.html#existence-of-objects)

But then you can express this as `record.users.exists?(user.id)`

PR #147, opened 2023-07-09
add in check in cta component

afogel:
This PR adds in the check in cta component. This CTA is meant to be a reasuable that directs user to check in if they haven't, edit their response if they have submitted a startup meeting, and edit their response if they skipped their startup meeting.

One caveat here -- this PR does not handle the empty state wherein an there is no standup meeting defined for the date in question. To enable this functionality would require handling how the `StandupMeeting::SkipsController` and `StandupMeeting::CompletionsController` deal with unpersisted `StandupMeeting`s, as currently rendering this component depends on `StandupMeeting::ActionButtonComponent`, which expects that the standup meeting passed to it is persisted.

Screen shots:

Screen Shot 2023-07-09 at 7 04 58 PM
Screen Shot 2023-07-09 at 7 04 40 PM
Screen Shot 2023-07-09 at 7 05 38 PM


app/controllers/standup_meetings_controller.rb
```diff
@@ -7,6 +7,10 @@ def index
     @standup_meetings = @standup_meeting_group.standup_meetings
                                               .includes(:user)
                                               .where(meeting_date: @meeting_date)
+    @current_user_standup_meeting = @standup_meetings.find do |meeting|
```
toyhammered, line 10: small nit, can you use `detect`. Unsure if this is defined in our rubocop rules but `find`/`where` etc I associate with ActiveRecord while `detect` and `select` are arrays.

toyhammered (approved):
small nit but otherwise this is awesome!

PR #150, opened 2023-07-10
User names in UI

### LINKED ISSUE #142: Replace user email with users' names in views
## Story
Currently, users are identified by their email addresses throughout the app. Examples can be seen when viewing pair requests or feedback:
Selection_203

With the completion of #115 , users would like to see others identified by name rather than email. The wireframes do a good job of showing how this should look: https://balsamiq.cloud/se3m7v2/puupn9o/r7BA4

## A PR Should
Make changes to to view templates and ViewComponents so that the user name is displayed on the following pages instead of their email:
- [x] pair request index
  - [x] Be sure to include the invitee select field for the new pair request form on the index page. Its options should display users by their names rather than email
- [x] pair request show
- [x] feedback index
- [x] feedback show

JoshDevHub:
##  Description
- Display user names instead of user emails throughout the UI (ie for creating and reading pair requests and feedback)

## Screenshots
Selection_213

## Issue
Closes #142

### DIFF
--- a/app/views/pair_requests/_form.html.erb
+++ b/app/views/pair_requests/_form.html.erb
@@ -3,7 +3,7 @@
     <div id="form-container" class='flex flex-col space-y-4 xl:space-y-0 xl:max-w-full xl:inline-flex xl:justify-around xl:space-x-8 xl:flex-row'>
       <div class="flex justify-between space-x-4 align-middle">
         <%= form.label :invitee_id, class: "label label-text" %>
-        <%= form.collection_select :invitee_id, User.without(current_user), :id, :full_name, {}, { class: 'select select-primary' } %>
+        <%= form.collection_select :invitee_id, User.excluding(current_user), :id, :full_name, {}, { class: 'select select-primary' } %>
       </div>
 
       <div class="flex justify-between space-x-4 align-middle">
--- a/app/views/pair_requests/_form.html.erb
+++ b/app/views/pair_requests/_form.html.erb
@@ -3,10 +3,7 @@
     <div id="form-container" class='flex flex-col space-y-4 xl:space-y-0 xl:max-w-full xl:inline-flex xl:justify-around xl:space-x-8 xl:flex-row'>
       <div class="flex justify-between space-x-4 align-middle">
         <%= form.label :invitee_id, class: "label label-text" %>
-        <%= form.select :invitee_id, 
-                        User.invitee_select_for(current_user), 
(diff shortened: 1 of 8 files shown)


app/views/pair_requests/_form.html.erb
```diff
@@ -3,10 +3,7 @@
     <div id="form-container" class='flex flex-col space-y-4 xl:space-y-0 xl:max-w-full xl:inline-flex xl:justify-around xl:space-x-8 xl:flex-row'>
       <div class="flex justify-between space-x-4 align-middle">
         <%= form.label :invitee_id, class: "label label-text" %>
-        <%= form.select :invitee_id, 
-                        User.invitee_select_for(current_user), 
-                        {}, 
-                        { class: 'select select-primary' } %>
+        <%= form.collection_select :invitee_id, User.all, :id, :full_name, {}, { class: 'select select-primary' } %>
```
JoshDevHub, line 6: Forgot about this form helper when I initially worked on this code. This obsoletes the need for the `invitee_select_for` scope that we were previously using here.

app/views/pair_requests/_table_row.html.erb
```diff
@@ -1,6 +1,8 @@
 <div class="table-row odd:bg-white even:bg-neutral-100 hover:bg-neutral-200 divide-x-2 divide-black">
   <%= render TableCellComponent.new(container_link: pair_request) do %>
-    <%= pair_request.partner_for(current_user).email %>
+    <% partner = pair_request.partner_for(current_user) %>
+    <p><%= partner.first_name %></p>
+    <p><%= partner.last_name %></p>
   <% end %>
```
toyhammered, line 6: should we not use `full_name` here?

app/views/pair_requests/_form.html.erb
JoshDevHub: Actually just thought of a small bug with this ... `User.all` will include the current user. Which probably isn't a great idea. I'll push up a fix.

toyhammered (approved):
One small comment but looks good otherwise.

app/views/pair_requests/_table_row.html.erb
JoshDevHub: These are intentionally placed in separate block level elements so they are rendered to HTML on separate lines:
Selection_214
For the table specifically, I think this looks good, especially on mobile. It also matches the way they're done in the wireframes.

PR #153, opened 2023-07-10
151 new standups

### LINKED ISSUE #151: [Standup Epic] Create New StandupMeeting when not already created by cron job
Currently, the system expects that we are editing existing StandupMeetings created by a cron job for every member of a standup meeting group. The work included in this PR should modify the existing flow to enable the creation of new `StandupMeeting`s in the event that one does not exist for a particular user in a group, then redirect to the correct subsequent action.

afogel:
closes #151 .

Please review after https://github.com/agency-of-learning/PairApp/pull/147 as this PR is based off of that branch.

In this PR, we add a create action in the event that a standupmeeting is not present for the user for a given date. The expected behavior here is that when there is no standup meeting, a user is provided the opportunity to check in.

Clicking on the "check in" button will create a new check in and redirect to the edit form that already exists.

I omitted the ability for the user to skip a check in when it's not created for two reasons: 
1) Presumably, if someone wants to check in for a date in which they don't already have a standup meeting, they likely only want to add standup information, not skip information.
2) the implementation of the skips controller requires a post action. Part of the HTTP spec prohibits redirections to be anything beyond GET methods. To maintain parsimonious controller logic and not introduce an ugly hack around the HTTP spec, I omitted the ability to skip if the StandupMeeting is not persisted.


app/components/standup_meeting/check_in_cta_component.html.erb
```diff
@@ -17,13 +17,22 @@
...
         current_user:) %>
-      <%= link_to "Check in", edit_standup_meeting_group_standup_meeting_path(standup_meeting_group, standup_meeting), class: 'btn btn-primary' %>
+      <% if standup_meeting.persisted? %>
+        <%= link_to "Check in", 
+                    edit_standup_meeting_group_standup_meeting_path(standup_meeting_group, standup_meeting), 
+                    class: 'btn btn-primary' %>
+      <% else %>
+        <%= button_to "Check in", 
+                      standup_meeting_group_standup_meetings_path(standup_meeting_group, meeting_date: standup_meeting.meeting_date), 
+                      class: 'btn btn-primary' %>
```
toyhammered, line 35: should this potentially be a component? (would also remove the nested if)

toyhammered (commented):
This is great! One small comment but I am super stoked to get this live and start testing because I think it is almost at that point (especially since this allows for us to create standups w/o the need of a cron job).

app/components/standup_meeting/check_in_cta_component.html.erb
toyhammered: thanks for taking the time to think/write about this. I am totally cool with what you said and it makes sense.

toyhammered (approved):

Issue #157, opened 2023-07-11
Style StandupMeeting::MeetingUpdateComponent rendered content 


GALTdea:
Currently  the StandupMeeting::MeetingUpdateComponent in standup_meeting index view looks like this:

<img width="1146" alt="Screenshot_2023-07-11_at_9 56 51_AM" src="https://github.com/agency-of-learning/PairApp/assets/16809030/737bf0f4-1b1c-474b-9c02-96e904054aeb">

This issue requires the content in each of the rows to display the user_name like this:

Daniel Rassiner
(grey line to separate or something)
Content here

PR #159, opened 2023-07-12
#157 style meeting update component conten

GALTdea:
I have updated the code to render the user's name above the content. However, I am submitting a draft Pull Request because there is an issue with the columns. Specifically, when one of the columns has no content, it fails to maintain its width. 

 
Screen Shot 2023-07-11 at 9 08 42 PM

with daisyui line
Screen Shot 2023-07-13 at 3 39 41 PM

### DIFF
--- a/app/components/standup_meeting/meeting_update_component.html.erb
+++ b/app/components/standup_meeting/meeting_update_component.html.erb
@@ -1,10 +1,9 @@
 <div class="border p-4 flex flex-col  justify-between mt-4">
-  <div class="w-full mb-1 font-bold justify-center">
-    <%= user.email %>
-  </div>
-  <div class="grow w-full">
+  <span class="flex flex-col justify-center mr-4 font-bold">
+    <%= user.full_name %>
+  </span>
+  <hr class="border-t border-gray-400 my-4" />
+  <span class="grow">
     <%= standup_meeting.public_send(content_type) %>
-  </div>
+  </span>
 </div>
-
-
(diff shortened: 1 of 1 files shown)


toyhammered:
Screenshot 2023-07-11 at 9 41 39 PM
I just pulled it down and it seems ok?

app/components/standup_meeting/meeting_update_component.html.erb
```diff
@@ -1,5 +1,8 @@
-<div class="border p-4 flex flex-col lg:flex-row justify-between mt-4">
-  <span class="flex flex-col justify-center mr-4 font-bold"><%= user.email %></span>
+<div class="border p-4 flex flex-col  justify-between mt-4">
+  <span class="flex flex-col justify-center mr-4 font-bold">
+    <%= user.full_name %>
+  </span>
+  <hr class="border-t border-gray-400 my-4" />
```
afogel, line 5: nit, but so you should know -- you can use the native daisy class:
`<div class="divider my-4 "></div>`

afogel (commented):
Overall, lgtm! re: formatting, as daniel mentioned, there looks to be an issue on your local machine re: daisy styling. I'd try to clobber your assets, make sure you're yarn installed and bundled, restart your server, and see whether that improves things.

app/components/standup_meeting/meeting_update_component.html.erb
GALTdea: done

GALTdea:
ok added the daisyui line? let me know what you think @toyhammered @afogel

toyhammered (approved):

PR #161, opened 2023-07-12
Fix navbar wrapping

### LINKED ISSUE #143: Navbar links wrapping to newline
The navbar links on the right side of the app's header are much too eager to wrap to a newline, which doesn't look great.

**Image**
Selection_205

The goal here should be to keep these links inline without wrapping.

## A PR Should:
- [x] Make sure the links stay inline at wide views.
- [x] Include a screenshot of the navbar at ~1050px width (just before the breakpoint to hamburger nav)

- If things start to look too cramped, we can think about bringing in the hamburger navbar for wider viewports. Right now the hamburger nav is there up until the Tailwind `lg` breakpoint(1024px).
- Make sure you look at how things appear to an admin user. They have an extra link in the navbar for inviting new users to the app.

JoshDevHub:
## Description
- Fix navbar wrapping
- Small cleanup/refactor of some markup (redundant `get` request options on a `link_to`, unneeded classes/styles, etc.)

#### 1024px
Selection_216

This is what it looks like most cramped at 1024px width. When less than this, the hamburger nav will display instead.

## Issue
Closes #143

### DIFF
--- a/app/views/layouts/_navbar.html.erb
+++ b/app/views/layouts/_navbar.html.erb
@@ -1,16 +1,16 @@
-<nav class="navbar bg-base-100 p-4">
-  <div class="navbar-start">
+<nav class="navbar bg-base-100 p-4 flex justify-between">
+  <div>
     <%= link_to "Agency of Learning", root_path, class: "text-xl" %>
   </div>
-  <div class="navbar-end">
+  <div>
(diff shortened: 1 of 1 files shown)


afogel (approved):
lgtm shippit

PR #162, opened 2023-07-13
enforce a date param, and that it's always passed in the correct format

afogel:
So this isn't my favorite looking code, but I think doing this:
1. enforces that there's always a date parameter
2. enforces that the date parameter is always formatted correctly


toyhammered (approved):
fine for now can fix up later

PR #171, opened 2023-07-17
fix bug in meeting when checked-in in another group

afogel:
Surfaced by Shami -- fixes bug that shows that you've already completed a check in, even though it was in a different group.

Bug was that the find_by didn't also query on the group. This fixes the bug.

To test:
check out branch, boot up server, replant your seeds, and log-in as user. Create a second standup meeting group.
navigate to the show page for one of the groups, and skip the meeting.
Navigate to the show page of the second group. Your check in for the day should not show that you skipped the standup meeting.


ShamiTomita (approved):
Awesome!

PR #172, opened 2023-07-18
Split workers from jobs

JoshDevHub:
## Description
- Move current jobs (which use sidekiq API) to new `workers/` folder and change their names to use `Worker`
  - This includes a small failsafe to prevent currently enqueued `AutoExpireWorkers` with their `Job` name from failing
- Create new `ApplicationJob` that inherits from `ActiveJob`
- Configure `ActiveJob` to use the sidekiq queue adapter


afogel (approved):
lgtm

PR #174, opened 2023-07-20
Notification to remind users to complete pair requests

JoshDevHub:
## Description
- Adds reminder notification for completing pair requests
- Adds mailer for this notification

### DIFF
--- a/app/controllers/pair_requests/acceptances_controller.rb
+++ b/app/controllers/pair_requests/acceptances_controller.rb
@@ -3,9 +3,16 @@ class PairRequests::AcceptancesController < ApplicationController
     @pair_request = PairRequest.find(params[:pair_request_id])
     authorize @pair_request, policy_class: PairRequest::AcceptancePolicy
     @pair_request.accepted!
-    PairRequest::AcceptanceNotification.with(pair_request: @pair_request).deliver(@pair_request.author)
+    send_notifications
 
     flash[:notice] = 'You have accepted this pair request.'
     redirect_to pair_request_path(@pair_request)
+  end
+
+  private
+
+  def send_notifications
+    PairRequest::AcceptanceNotification.with(pair_request: @pair_request).deliver(@pair_request.author)
+    PairRequest::CompletionReminderNotification.with(pair_request: @pair_request).deliver(@pair_request.author)
   end
 end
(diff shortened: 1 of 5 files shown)


app/notifications/pair_request/completion_reminder_notification.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  private
+
+  def now_to_end_of_meeting
+    end_time = params[:pair_request].end_time
+    ActiveSupport::Duration.build(end_time - Time.current)
+  end
+
+  def pair_request_accepted?
+    params[:pair_request].accepted?
```
JoshDevHub, line 18: Intentionally checking for the pair request being in an `accepted` state here. If I were to check for "not complete", it could open up edge case issues with pair requests that have been canceled or rejected.

app/controllers/pair_requests/acceptances_controller.rb
```diff
@@ -3,9 +3,16 @@ def create
...
 
     flash[:notice] = 'You have accepted this pair request.'
     redirect_to pair_request_path(@pair_request)
   end
+
+  private
+
+  def send_notifications
+    PairRequest::AcceptanceNotification.with(pair_request: @pair_request).deliver(@pair_request.author)
+    PairRequest::CompletionReminderNotification.with(pair_request: @pair_request).deliver(@pair_request.author)
```
toyhammered, line 16: Just one question related to this. It seems like there is a delay set. Is the `if: :pair_request_accepted?` only triggered once it trys to send the email?

So the flow is:
1. invitee accepts pair request
2. creates the send notifcations for the author
3. the completion one is only triggered once it is after the delay set. If at this point the pair request is still in an acceptance, it will send a reminder, otherwise it will just noop?

toyhammered (commented):
This is great! Just have 1 clarifying questions before I approve.

app/controllers/pair_requests/acceptances_controller.rb
JoshDevHub: Yes exactly. The condition is used when it comes time to send the email. If the condition is false when it's time to send it, nothing happens. So users who have already marked the request as completed (or they've canceled or whatever) won't get the email.

JoshDevHub: I was initially a bit confused about this too but parts  of the Noticed doc make it clear what's happening. Like [here in the README](https://github.com/excid3/noticed/tree/master#fallback-notifications) he shows a pattern for fallback notifications that also depends on the conditions being evaluated on a delay.

toyhammered (approved):

PR #190, opened 2023-07-22
Add eslint for linting JS

### LINKED ISSUE #168: Linter for JS
**Is your feature request related to a problem? Please describe.**
I see inconsistencies in linting/style in JS. So far, it's not being addressed, but I think it provides consistency for the entire team.

**Describe the solution you'd like**
Add a linter. [Eslint](https://eslint.org/) has fairly broad adoption, so might be a good option.

**Describe alternatives you've considered**
A cursory [look at some other options](https://blog.logrocket.com/modern-faster-alternatives-eslint/) suggests that there are alternatives. Rome looked most interesting to me, but we'll see.

JoshDevHub:
## Description
- Add eslint for linting js files
- Lint existing javascript files
- Adds javascript linting to CI workflow

## Issue
Closes #168 

## Eslint Config
I used the "airbnb" config mostly because it's one I've used and liked in the past. Didn't think there would be too many strong opinions about what we go with here. It kinda is necessary to use some form of outside config because eslint on its own has *very* few opinions on style/format and setting up the rules manually would be a pain.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,12 @@
+{
+  "env": {
+    "browser": true,
+    "commonjs": true,
+    "es2021": true
+  },
+  "extends": "airbnb-base",
+  "parserOptions": {
+    "ecmaVersion": "latest",
+    "sourceType": "module"
+  }
+}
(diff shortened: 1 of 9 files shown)


afogel (approved):
awesome! shipit!

PR #191, opened 2023-07-22
Faq page

### LINKED ISSUE #98: FAQ Page
Create a simple FAQ Page that will give some helpful tips to the people about this app, purpose, and common questions they might have. Some starters can be:
1. What is the purpose of this app?
2. Who would benefit from this app?
3. How do I contribute to this app?

This page does not need to be supported by a model. All these questions/answers can be hardcoded in the front-end (or as a const in the controller). For the design doc, the only thing that it needs to include is a place where we can disucss which questions to add.

The wireframes for this one are important just so we can make this feel right. (This can be done on balsamiq, if you need an invite let @toyhammered know)

afogel:
## What's the change?
Add a FAQ. Also, I was surprised when the PR template didn't auto trigger, so I'm trying to move it to a different folder to see whether that signals to the github API to use our PR template by default.

To test, you should open up the pair app locally and make sure you're not logged in. You should see a FAQ link in the navbar. Clicking it will take you to the FAQ.
If you log in, you should not see the FAQ link in the nav bar.
No matter what, you should be able to access the FAQ, though, by navigating to `/faq`.

Adds a static controller that enables static content to be served without authentication. 

## Demo / Screenshots

Screen Shot 2023-07-22 at 9 18 13 AM
Screen Shot 2023-07-22 at 9 18 04 AM
Screen Shot 2023-07-22 at 9 17 53 AM

Closes #98 

- [X] Did you include instructions for how to test in the ticket?


JoshDevHub (approved):
LGTM :rocket:

PR #193, opened 2023-07-25
Setup ActiveStorage for profile pictures

JoshDevHub:
## What's the change?
- Generate and run migrations for ActiveStorage blobs
- Add `active_storage_validations` gem for validating ActiveStorage attachments
- Allow profiles to attach a `:picture` image
- Build very basic profile show page so devs can see a profile picture displayed

## Testing

I have a very basic show and edit/update flow in the front end. This exists to help with testing this feature locally and in production.

## TODO

So this is in draft. The big thing to figure out from here is getting this hooked up in production. This should be pretty straightforward with an S3 bucket. I don't have the keys on that though. You should just have to comment out [this section](https://github.com/agency-of-learning/PairApp/blob/main/config/storage.yml#L10-L15) of `config/storage.yml` and supply valid credentials and install the [`aws-sdk` gem](https://github.com/aws/aws-sdk-ruby).

Another thing possibly worth discussing is that while we iterate on the profile stuff, we probably don't want to actually have the pieces of this feature fully live in prod until there's something substantial for the user to work with. We can just make it opaque by not linking to it anywhere and that's probably fine for our scale and usecase right now. Or we could also explore feature flagging to strictly hide WIP stuff.


JoshDevHub:
S3 configs are pushed up and should be good to go.

To test this in production, you'll have to learn the id of a user's profile. From their, you can visit `/profiles/:id` to go to that user's profile. Click "Edit Profile" and you'll get the form where you can upload an image. The `profiles#show` page should display this image after uploading.

toyhammered (approved):
This is great! I think we should get prod keys first but not really a blocker.

toyhammered (approved):

PR #194, opened 2023-07-25
Allow standup content to render newline characters

JoshDevHub:
## What's the change?
- Allow newline characters to be rendered for standup descriptions.
- Small refactor to extract template logic in a component (leave code better than you found it, etc.)

Submitting and reading standup meetings.

When submitting standup, users can now freely add newlines with Enter/Return to visually organize their update. This will be reflected in the meeting output.

#### What it looks like now:
Selection_223

#### What it did look like:
Selection_224

Edit:
- [x] Add automated tests -- I had incorrectly assumed the component I touched already had coverage.


toyhammered (approved):

afogel:
This is 🔥 !

Issue #152, opened 2023-07-10
[Standup Epic] Cron job to create new standup instances for members of standup meetng groups


afogel:

toyhammered:
I'm wondering if we want to put this together with integrating with notifier gem. That is the real value in having these go out as cron jobs.

afogel:
Now that we have the ability to create a new standup meeting, even on days when we don't have existing instances, I'd agree. That said, it could be an easy first issue for someone unfamiliar with writing cron tasks to try out, and could be augmented with the new functionality that is brought in by the notifier gem once we integrate it. I don't feel too strongly about this either way, but I figured making an issue was worthwhile so that it wouldn't fall off the radar.
What do you think?

toyhammered:
I think if we implement the notifier gem ourselves as a separate task and include an example (for like Pair Request) then yeah this could be good. There are some nuances for when these should be created but could help them along with it.

toyhammered:
Now that the notification gem is implemented, I'll quickly knock this out.

PR #195, opened 2023-07-26
Add notification for standup meetings

toyhammered:
## What's the change?
Adding a cronjob to send out notifications for standups that are upcoming and you haven't responded to yet.

https://github.com/agency-of-learning/PairApp/issues/152


app/workers/standup_meeting_group/determine_missing_standup_meetings_worker.rb
```diff
@@ -0,0 +1,45 @@
...
+                             end_time
+                           )
+                           .where('standup_meetings.id IS NULL OR standup_meetings.meeting_date = ?', current_date)
+                           .group('standup_meeting_groups.id')
+                           .having('COUNT(standup_meeting_groups_users.id) > COUNT(DISTINCT standup_meetings.id)')
+
+      # NOTE: later on if this becomes a bottle neck we can have a separate worker
+      # for each standup_meeting_group (and pass in user_ids).
+      # Ideally though we would just batch these at that point.
+      standup_meeting_groups.each do |standup_meeting_group|
```
afogel, line 34: That makes sense to me -- I think we're going to have very few standup meeting groups, overall, so despite the O(N^2), I'd say it's probably good enough. I guess we'll just monitor sidekiq for failing jobs to identify if/when this becomes problematic?

app/workers/standup_meeting/create_draft_worker.rb
```diff
@@ -0,0 +1,19 @@
...
+        standup_meeting_group_id:, user_id:, meeting_date: date
+      )
+
+      return if standup_meeting.persisted?
+
+      standup_meeting.save!
+
+      user = standup_meeting.user
+
+      StandupMeeting::DraftCreationNotification.with(standup_meeting:).deliver(user)
```
afogel, line 16: I think we want to `deliver_later` instead of `deliver` so that this work gets unqueued and further distribute the blocking processes?

afogel (changes requested):
overall, looks good! One small thing to update in the sidekiq `CreateDraftWorker` where I think we can enqueue the mail delivery to make the `CreateDraftWorker` simpler/more efficient, especially since mail delivery is not time sensitive.

afogel (approved):
Looks great!

Issue #192, opened 2023-07-24
Incorrect error message when creating a future pair request


RocDavenport:
**Describe the bug**
When attempting to create a Pair Request dated beyond 1 month in the future, the error message states "Error! Please fix inputs When field must not be in the past"

**To Reproduce**
1. Go to 'Pair Requests'
2. Select an 'Invitee'
3. Enter a 'Date' more than a month in the future.
4. Select a 'Duration'
5. Select an 'Create Pair Request
6. See error

**Expected behavior**
I expect to see an error message that is applicable to the date selected.

**Screenshots**
<img width="1277" alt="AOL_my_request_errorpng" src="https://github.com/agency-of-learning/PairApp/assets/134003653/c2db42ce-448f-4ed8-b0e1-cc5941621897">

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser Chrome 
 - Version 114

**Additional context**

GALTdea:
I think you should tackle this one @RocDavenport.

afogel:
I agree that the error message could probably be improved, but might this be an issue with the fact that the invitee isn't selected there?

RocDavenport:
No I think what happened here is once the error was thrown, the invitee
field was defaulted back to no one. The screenshot was captured after the
failed attempt to schedule a pair request.

RocDavenport:
And so it is! Thank you!

afogel:
@RocDavenport wait, does that mean that the cause was that invitee was unselected?

RocDavenport:
@Ariel no, and I apologize for the confusion.

When I attempted to schedule the Pair Request, I received the error
message. At the time the attempt was made, all fields were completed.

I believe the error stems from the fact that I accidentally entered a date
in September (more than 1 month away).

When I then took the screenshot, it appeared as if no invitee was selected
at the time of the pair request attempt, but that is not the case.

afogel:
@RocDavenport got it, great catch! Looking forward to seeing what the fix is 😎

PR #196, opened 2023-07-28
Update the Pair Request Error message

RocDavenport:
## What's the change? 
Changed the error message to reflect the two possible error cases.
1. When the pair request date is in the past
2. When the pair request date is more than 1 month in the future.

None

This change should be temporary, until a more appropriate fix can be completed.

## Demo / Screenshots
Screen Shot 2023-07-28 at 2 15 17 AM

#192 https://github.com/agency-of-learning/PairApp/issues/192
n/a


toyhammered (approved):
Thank you!

PR #197, opened 2023-07-28
New profile fields

JoshDevHub:
## What's the change?
- Add new fields for profiles:
    - location
    - job_search_status
    - work_model_preferences

The `work_model_preferences` field uses a Postgres array typed with a Postgres enum. This shouldn't be too weird to work with when we add some helpers and scopes for working with the data. I was not able to figure out a way to add a default of an empty array, which is a bit annoying. I welcome ideas on that.

## Draft Notes

Throwing this in draft for now so people can take a look. The main thing I don't really like is that I couldn't find a way to write a database default for the `work_model_preferences` field. I don't think it'll be too bad to work around though.

I was able to confirm that this works on migrate and on schema load (the latter can sometimes be picky when migrations run raw SQL). It can also be rolled back with no problems.

As far as logic around the `work_preferences_field`, I've just written basic common sense validations for now (can't insert a key that's not in the enum, can't add duplicate keys). It will definitely be nice to have some helpers, scopes, etc. to work with the data, but I think I'll leave that to PRs that actually need that rather than trying to anticipate what we'll need. I do still want to write tests for the validation logic.


app/models/profile.rb
```diff
@@ -19,4 +22,32 @@
 #
 class Profile < ApplicationRecord
   belongs_to :user
+
+  enum job_search_status: {
```
toyhammered, line 26: very small thing, i'd make `not_looking` as 0 and `actively_looking` as 2.

app/models/profile.rb
```diff
@@ -19,4 +22,32 @@
...
+    actively_looking: 0,
+    open_to_opportunities: 1,
+    not_looking: 2
+  }
+
+  WORK_MODELS = %w[inoffice hybrid remote].freeze
```
toyhammered, line 32: instead of `inoffice` can we do `onsite`

app/models/profile.rb
```diff
@@ -19,4 +22,32 @@
...
+    if work_model_preferences.any? { |preference| WORK_MODELS.exclude?(preference) }
+      error_message = "must only include valid preferences: #{WORK_MODELS.to_sentence}"
+      errors.add(:work_model_preferences, error_message)
+    end
+  end
+
+  def must_not_duplicate_preferences
+    if work_model_preferences.size != work_model_preferences.uniq.size
+      errors.add(:work_model_preferences, 'must not contain duplicate preferences')
+    end
```
toyhammered, line 51: small ruby thing, we generally prefer doing guard clauses over if statements + method finish. Sometimes it isn't possible though.

For `must_not_duplicate_preference` we can easily turn that into `return if work_model_preferences.size == work_model_preferences.uniq.size`

toyhammered (changes requested):
This looks great, couple small things but great job otherwise!

.rubocop.yml
```diff
@@ -324,7 +324,7 @@ RSpec/MultipleMemoizedHelpers:
   Enabled: false
 RSpec/NestedGroups:
   Enabled: false
-Rspec/MessageSpies:
+RSpec/MessageSpies:
```
JoshDevHub, line 327: rubocop was very mildly complaining about the casing here in its output, so I went ahead and fixed it.

app/workers/standup_meeting_group/determine_missing_standup_meetings_worker.rb
```diff
@@ -13,7 +13,11 @@ def perform(start_minutes = 0, end_minutes = 0)
 
       standup_meeting_groups =
         StandupMeetingGroup
-        .joins("LEFT OUTER JOIN standup_meetings ON standup_meeting_groups.id = standup_meetings.standup_meeting_group_id AND standup_meetings.meeting_date = '#{current_date}'")
+        .joins(
+          "LEFT OUTER JOIN standup_meetings
+            ON standup_meeting_groups.id = standup_meetings.standup_meeting_group_id
+            AND standup_meetings.meeting_date = '#{current_date}'"
+        )
```
JoshDevHub, line 20: rubocop didn't like this line length. I couldn't pass CI without fixing it.

toyhammered: oh thats on me :D. I should have just disabled for that line so ty.

toyhammered (approved):

PR #205, opened 2023-07-31
Add Noticed Notification table

toyhammered:
## What's the change?
Adds the table for Noticed gem so that specific Notifications get stored when created.

### DIFF
--- a/app/models/pair_request.rb
+++ b/app/models/pair_request.rb
@@ -22,9 +22,11 @@
 #  fk_rails_...  (invitee_id => users.id)
 #
 class PairRequest < ApplicationRecord
+  # has_many :notifications, as: :recipient, dependent: :destroy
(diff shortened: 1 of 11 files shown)


app/models/pair_request.rb
```diff
@@ -22,9 +22,11 @@
 #  fk_rails_...  (invitee_id => users.id)
 #
 class PairRequest < ApplicationRecord
+  # has_many :notifications, as: :recipient, dependent: :destroy
```
JoshDevHub, line 25: This could be removed I think?

JoshDevHub (approved):
Just one small, non-blocking comment. Looks good :rocket:

app/models/pair_request.rb
toyhammered: yes, forgot to remove that ty.

PR #206, opened 2023-08-01
Add letter opener web for easier management of emails

toyhammered:
## What's the change?
Added a gem to give us a better interface for working with emails in dev. When we original setup `letter_opener` something felt off and this was the reason.


JoshDevHub (approved):
Good stuff!

Issue #182, opened 2023-07-20
[Profiles] Make profile public/private


afogel:
Blocked until #179 , #180, and #181 are finished.
Adds ability to make accessing the custom profile public or private.

- [ ] When Private, user without permission should get redirected
- [ ] When Private, users with permission should be shown profile
- [ ] When Private, profile owner (and only profile owner) should be shown a button to toggle profile from private to public
- [ ] When Public, all users should be shown
- [ ] When Public, profile owner (and only profile owner) should be shown a button to toggle profile from public to private

toyhammered:
> Blocked until #179 , #180, and #181 are finished. Adds ability to make accessing the custom profile public or private.
> 
> * [ ]  When Private, users with permission should be shown profile

I'd like to keep this even simpler. When private, only owner and admin can see. No need for any permissions or anything initially.

PR #207, opened 2023-08-01
Profile show page

### LINKED ISSUE #180: [Profiles] Profiles#Show page
1. Space for resume upload/show (functionality can be added in #179 )
2. Profile picture
3. short bio or description
  i. Add a new column to the user table that includes this. We can consider a new table that stores more of this info separately purely so the user table doesn’t become over bloated. Figuring out where that fine line is, is something that will need to be explored (and what fields we want where).
This could be in something potentially called a “profile”.

JoshDevHub:
## What's the change?
- Style currently existing pieces of the profile show page
- Create placeholder edit form for profiles to make this feature easier to work with
- Create pundit policy for profiles
    - A bit of a placeholder. Will expect it to change a lot when #182 is done
- Add component for rendering the user's profile picture
- Add placeholder profile image for when users have not uploaded a profile picture
- Setup testing around file attachments

#### Mobile
Selection_231

#### Desktop
Selection_230

Closes #180 

## Misc
This is still opaque to the user (they'd have to know the route and know a profile ID to visit it in production), so users still can't use this in prod. We could link to it in the navbar and tell people to start using it, or we could wait until the feature is more polished and complete.


app/components/profile/picture_component.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class Profile::PictureComponent < ViewComponent::Base
+  JOB_SEARCH_COLOR_CLASSES = {
+    'not_job_searching' => 'bg-red-800 text-red-50',
+    'open_to_opportunities' => 'bg-amber-800 text-amber-50',
+    'open_to_work' => 'bg-green-800 text-green-50'
```
JoshDevHub, line 7: Intentionally not being as "programmatic" about this. Tailwind needs the full class name for each utility or else that utility will be caught in its "purging" -- where Tailwind goes through and removes unused styles as a way to control bundle size.

app/models/profile.rb
```diff
@@ -42,6 +42,10 @@ def work_model_preferences
     super || []
   end
 
+  def work_model_preferences=(values)
+    super(values.reject(&:empty?))
+  end
```
JoshDevHub, line 47: Felt this was the easiest way to deal with how the checkboxes submit their data. If the user checks "remote" only, it'll come in as `["", "", "remote"]`.

app/models/profile.rb
```diff
@@ -30,9 +30,9 @@ class Profile < ApplicationRecord
   validates :picture, content_type: ['image/png', 'image/jpeg']
 
   enum job_search_status: {
-    not_looking: 0,
+    not_job_searching: 0,
     open_to_opportunities: 1,
-    actively_looking: 2
+    open_to_work: 2
```
JoshDevHub, line 35: Changed these because when actually viewing the profile, it felt strange to read things like "not actively looking". Not actively looking for what?

Assuming that people outside the agency may eventually see these pages, it's probably good to explicitly say things about "work", "job", etc. in these so it's very clear what's being communicated.

Also definitely don't consider what I have here the "best" and am interested on feedback. This naming is tricky to get right.

spec/factories/profiles.rb
```diff
@@ -26,7 +26,21 @@
...
+    end
+
+    after(:build) do |profile, evaluator|
+      next unless evaluator.attached_picture
+
+      profile.picture.attach(
+        io: Rails.root.join('spec/fixtures/test_image.png').open,
+        filename: 'test_image.png',
+        content_type: 'image/png'
+      )
```
JoshDevHub, line 43: This is something I'm less familiar with and definitely welcome feedback if there's a better way to work with attachments in tests and with factorybot.

afogel (approved):
overall, this looks 🔥 . Killing it. I'm not sure about the fixture thing, either, but it makes sense to me how you pulled it together.

spec/factories/profiles.rb
toyhammered: it's been a hot minute since i've down image tests lol. If this work, its fine for now we can fine tune it later.

Issue #204, opened 2023-07-30
Check-in Status Messages in Joined Card Components


GALTdea:
### Description
As a user, I should be able to see a  check-in status message in my joined_card_components indicating weather I've completed the standup or not.   Currently, the joined_card_components display the necessary HTML, styling, messaging, and icons. However, there is a need to add view components to make this messages dynamic. 

### steps
1. Informative Status Display: Embed view components within the joined_card_components to display standup check-in status messages prominently. This feature will enable users to quickly grasp their daily standup completion status.

PR #209, opened 2023-08-03
 Checkin status messages

GALTdea:
## What's the change?
A new ViewComponent named StandupMeetingGroup::CheckInStatusComponent 
A specs for the new view component
Modifications in the standup group meeting index action to eager load standup_meetings

- It Allows users to visualize their joined standup_meeting_groups checkin status from the index page
-
Check-in Status Messages in Joined Card Components #204

### DIFF
--- a/app/components/standup_meeting_group/check_in_status_component.html.erb
+++ b/app/components/standup_meeting_group/check_in_status_component.html.erb
@@ -1,0 +1,5 @@
+<% if @checked_in %>
+  <p class="text-xs"> <i class="fa-solid fa-check"></i> Checked in </p> 
+  <% else %>
+  <p class="text-xs"> <i class="fa-solid fa-triangle-exclamation"></i> Outstanding check in</p> 
+<% end %>
--- a/app/components/standup_meeting_group/check_in_status_component.html.erb
+++ b/app/components/standup_meeting_group/check_in_status_component.html.erb
@@ -1,8 +1,5 @@
-<% if @standup_meeting.present? && @standup_meeting.completed? %>
+<% if @standup_meeting&.completed? %>
   <p class="text-xs"> <i class="fa-solid fa-check"></i> Checked in </p>
 <% else %>
   <p class="text-xs"> <i class="fa-solid fa-triangle-exclamation"></i> Outstanding check-in</p>
 <% end %>
-  
-
-
--- a/app/components/standup_meeting_group/check_in_status_component.html.erb
+++ b/app/components/standup_meeting_group/check_in_status_component.html.erb
@@ -1,5 +1,8 @@
-<% if @checked_in %>
-  <p class="text-xs"> <i class="fa-solid fa-check"></i> Checked in </p> 
-  <% else %>
-  <p class="text-xs"> <i class="fa-solid fa-triangle-exclamation"></i> Outstanding check in</p> 
+<% if @standup_meeting.present? && @standup_meeting.completed? %>
+  <p class="text-xs"> <i class="fa-solid fa-check"></i> Checked in </p>
+<% else %>
+  <p class="text-xs"> <i class="fa-solid fa-triangle-exclamation"></i> Outstanding check-in</p>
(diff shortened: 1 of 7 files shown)


app/controllers/standup_meeting_groups_controller.rb
GALTdea, line 1: Here I am eager loading standup_meetings to avoid making a database call from CheckInStatusComponent. But  I'm not sure that i was successful. I  don't see a difference in the logs when adding it.

toyhammered (commented):
This is looking great. I feel like it is almost done?

app/components/standup_meeting_group/check_in_status_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<% if @standup_meeting.present? && @standup_meeting.completed? %>
```
JoshDevHub, line 1: I might like condensing this `if` check down to `if @standup_meeting&.completed?`, but that's minor and not something you have to change.

JoshDevHub (changes requested):

app/controllers/standup_meeting_groups_controller.rb
JoshDevHub: I think this is because you end up looking for the meeting with `find_by`, which is one of those "always hit the database" methods.

If you really want to avoid making the call from the view, you could explore some way to get that data using `CollectionProxy` methods (this is what comes back from a `has_many` association like `@standup_group.standup_meetings). To me though, I'm probably fine to leave it alone because I doubt N+1 ever really hurts us here. Open to other opinions on that though.

spec/components/standup_meeting_group/check_in_status_component_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+RSpec.describe StandupMeetingGroup::CheckInStatusComponent, type: :component do
+  let(:standup_meeting_group) { create(:standup_meeting_group) }
+  let(:standup_meeting_group_2) { create(:standup_meeting_group) }
+  let(:user) { create(:user) }
+  let(:standup_meeting) { create(:standup_meeting, user: user, standup_meeting_group: standup_meeting_group, status: 'completed') }
+
+  it "renders 'Checked in' message when user has checked in" do
+    render_inline(described_class.new(standup_meeting_group, user))
+
+    expect(page).to have_css("p", text: "Checked in")
```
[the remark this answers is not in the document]
GALTdea: Good point! Agree.

app/controllers/standup_meeting_groups_controller.rb
GALTdea: Thanks for this feedback. I can try to figure out how to do it with collection proxy methods, but I tend to agree with you. I'll be ok leaving it the way it is.

app/components/standup_meeting_group/check_in_status_component.html.erb
GALTdea: I like your suggestion better!

spec/components/standup_meeting_group/check_in_status_component_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+RSpec.describe StandupMeetingGroup::CheckInStatusComponent, type: :component do
+  let(:user) { create(:user) }
+
+  let(:standup_meeting_group) do
+    create(:standup_meeting_group).tap do |group|
+      group.standup_meetings << create(:standup_meeting, user:, status: 'completed')
+    end
+  end
+  let(:standup_meeting_group2) { create(:standup_meeting_group) }
```
JoshDevHub, line 13: I would maybe name these standup meeting groups differently. Maybe something like `standup_meeting_group_with_checkin` and `standup_meeting_group_no_checkin` or something. Just something that gives more information about how they're different at a glance.

This isn't really a blocker though

JoshDevHub (approved):
Just one small non-blocking comment. Feel free to merge.

PR #213, opened 2023-08-09
bug: fix seeds

crespire:
* Change `Time.current` to `Time.now`
* Update `completed_pair_request_data`'s `when` field to be in the future


crespire:
This can be closed, as the issue is fixed in #214

Issue #201, opened 2023-07-30
Implement TimeCop to fix Intermittent failing of Tests


toyhammered:
**Describe the bug**
We currently do not implement TimeCop (https://github.com/travisjeffery/timecop) so some of our tests which rely on being set to certain times can intermittently fail (specifically close to the next day mark, ~12am utc).

A good example would be `DetermineMissingStandupMeetingsWorkerSpec`.

**To Reproduce**

It is kind of a pain to reproduce, might be able to do so with mocking `Time.current`, but if the `start_time` and `end_time` end up being on different days the test will fail.

**Expected behavior**
We should implement TimeCop so that these don't become an issue in the future.

fpsvogel:
@toyhammered Another option, if we want to avoid adding a gem, is `ActiveSupport::Testing::TimeHelpers`: https://andycroll.com/ruby/replace-timecop-with-rails-time-helpers-in-rspec

PR #214, opened 2023-08-09
fix: Update pair request `when` validation

crespire:
* Added a conditional for the time box validation
* Updated spec to add test cases for new conditional validation

## What's the change?
Adds a conditional to the pair request `when` field validation. The validation only runs when the pair request status is `pending`

Pair request creation

- [x] Ran the linter 😅 

This PR may also impact #201

### DIFF
--- a/app/models/pair_request.rb
+++ b/app/models/pair_request.rb
@@ -34,8 +34,9 @@ class PairRequest < ApplicationRecord
 
   validates :when,
     presence: true,
-    inclusion: { in: (Date.current..(Date.current + 1.month)),
-                 message: 'field must not be in the past or more than 1 month into the future' }
+    inclusion: { in: (Date.current..(Date.current + 0.month)),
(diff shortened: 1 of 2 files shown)


app/models/pair_request.rb
```diff
@@ -34,8 +34,9 @@ class PairRequest < ApplicationRecord
 
   validates :when,
     presence: true,
-    inclusion: { in: (Date.current..(Date.current + 1.month)),
-                 message: 'field must not be in the past or more than 1 month into the future' }
+    inclusion: { in: (Date.current..(Date.current + 0.months)),
```
JoshDevHub, line 37: I assume the change to `0.months` from `1.month` was accidental? Seems to be the cause of the test failures in CI I think.

crespire: Good catch, yes. Will fix!

toyhammered (approved):
This looks good to me

Issue #181, opened 2023-07-20
[Profiles] Custom Url


afogel:
Links to the profiles should be custom routes, like linkedin offers. May consider using [friendly_id](https://github.com/norman/friendly_id) gem?

PR #215, opened 2023-08-09
feat: make profile slugged

crespire:
## What's the change?
* Install friendly_id
* Update profiles_controller to use friendly
* Update User model to update slug after update (callback)
* Update Profile model to use FriendlyID
* Add a spec for the profile controller
* Add a devise.rb support file for tests

User create (because of profile create callback)

#181 

- [x] Any validations to data needed? - Slugs need to be generated for all existing profiles: `Profile.find_each(&:save)` on prod console.

Rollback:
* Delete FriendlyID tables
* Remove slug column from profile
* Remove references to `friendly.find` in the `ProfilesController`


crespire:
I added a method `update_profile_slug!` to `User` which is not tested. I can add a quick test if required, but it is a private method which is why I didn't add a test initially.

app/models/user.rb
```diff
@@ -101,4 +102,11 @@ def time_zone_identifier
     member: 0,
     admin: 1
   }
+
+  private
+
+  def update_profile_slug!
```
toyhammered, line 108: kind of confused as to why we do this after update? Is this to force it to update the slug? Users should be able to choose their own slug and make sure it is uniq across all profiles (we shouldn't be doing any magic really).

Also for testing this because you did ask, you'd want to do something that forces this to be used (i.e: make an update, which would then trigger this, and check the record to see it is updated as expected)

config/initializers/friendly_id.rb
```diff
@@ -0,0 +1,107 @@
...
+  # calls to `Module.extend`.
+  #
+  # config.use :finders
+  #
+  # ## Slugs
+  #
+  # Most applications will use the :slugged module everywhere. If you wish
+  # to do so, uncomment the following line.
+  #
+  # config.use :slugged
```
toyhammered, line 54: should enable this

spec/requests/profiles_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.describe 'Profiles' do
```
toyhammered, line 3: Thank you for adding these. I still want to understand more what the difference between request specs and controller specs are because this is exactly how i'd do a controller spec.

toyhammered (commented):
Great start! Just a couple questions I have.

spec/requests/profiles_spec.rb
crespire: My understanding is that the RSpec team recommends folks phase out controller specs in favour of request specs, the primary reason being that behind the scenes a controller spec is more of a unit test while a request spec is more of an integration test and a "more realistic" testing scenario (according to the RSpec blog, anyway...).

I take a lot of queues from Jason Swett on this topic: https://www.codewithjason.com/use-controller-request-specs-rails-dont/

I've been writing a lot of request specs lately for reason 3 on my day job app, so just in the habit.

app/models/user.rb
GALTdea: not sure if Daniel would want it in this issue, but I love the idea of the "denylist of words" feat. Are there any gems/services that would help with that?

toyhammered: For the default, we can make it something automatically populated (i'd even say maybe `first_name.last_name` for the initial).

For characters we should allow, do something like all alphanumeric, period, hyphen and underscore (.-_). That should be good enough for now.

As for denylist, we should prob find a gem to help with that. For now we won't add anything.

spec/requests/profiles_spec.rb
toyhammered: Love it! Thanks for sharing

app/models/user.rb
crespire: > For the default, we can make it something automatically populated (i'd even say maybe `first_name.last_name` for the initial).
> 
> For characters we should allow, do something like all alphanumeric, period, hyphen and underscore (.-_). That should be good enough for now.
> 
> As for denylist, we should prob find a gem to help with that. For now we won't add anything.

In terms of allowed characters, I don't think we can allow `.` as that typically denotes a format in a request URL. How do you feel about alphanumeric with just `-_` as well?

crespire:
@toyhammered Comments addressed! I made the decision to drop the `.` character from the valid set as it would mess with format specification for URLs.

I also added the profile slug field to the form before the Bio text field, but happy to move that to a different order.

I toyed with displaying errors on the form, either via backend `<% if @profile.errors.any? %>` or a frontend Stimulus controller, but didn't pursue it. If you have any preference here, let me know and I can add in error display.

JoshDevHub:
@crespire I'll let Daniel look at this more in depth, but I can tell you a couple of things about the error handling for the profile edit page.

The pattern we've been using is to have the form errors in a flash message. You can see an example [here in the `PairRequestsController`](https://github.com/agency-of-learning/PairApp/blob/main/app/controllers/pair_requests_controller.rb#L33).

toyhammered (commented):
This looks great to me! You should handle form errors as @JoshDevHub suggested.

As for front-end validation, what type would you like to add (I am all for it)

crespire:
> This looks great to me! You should handle form errors as @JoshDevHub suggested.
> 
> As for front-end validation, what type would you like to add (I am all for it)

I was thinking of using HTML5 validation. For the slug field, add `required` and also `pattern: '^[a-zA-Z0-9_-]+$'` - this way the browser will prevent submission if it's empty or not the right format.

crespire:
I have added HTML5 validation and also updated the form to handle backend error display in the way the rest of the app does.

I didn't look closely at the style of some other pages to see how supplemental info is presented, but it should be a quick fix if we want to keep this approach. Currently, the message is "Please match the format requested." We can do something custom, but that'd require more JS.

toyhammered (approved):
Great job!

db/migrate/20230809011403_create_friendly_id_slugs.rb
```diff
@@ -0,0 +1,21 @@
+MIGRATION_CLASS =
+  if ActiveRecord::VERSION::MAJOR >= 5
+    ActiveRecord::Migration["#{ActiveRecord::VERSION::MAJOR}.#{ActiveRecord::VERSION::MINOR}"]
```
afogel, line 3: I'm curious -- can you say more about this MIGRATION_CLASS?

crespire: > I'm curious -- can you say more about this MIGRATION_CLASS?

This was a generated file, so I'm not able to provide any insight beyond a hunch that it's trying to support older Rails versions.

PR #216, opened 2023-08-09
Add UserMenteeApplication table + model

toyhammered:
## What's the change?
Added the UserMenteeApplication table + model. The resume will be added at a later stage.

### DIFF
--- a/db/migrate/20230809025152_create_user_mentee_application.rb
+++ b/db/migrate/20230809025152_create_user_mentee_application.rb
@@ -1,0 +1,23 @@
+class CreateUserMenteeApplication < ActiveRecord::Migration[7.0]
+  def change
+    create_table :user_mentee_applications do |t|
+      t.references :user, null: false, foreign_key: true, index: true
+
+      t.string :city, null: false
+      t.string :state, null: false
+      t.string :country, null: false
+
+      t.text :reason_for_applying, null: false
+      t.string :linkedin_url
+      t.string :github_url
+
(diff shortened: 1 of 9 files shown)


afogel (approved):

db/migrate/20230809025152_create_user_mentee_application.rb
```diff
@@ -0,0 +1,23 @@
...
+  def change
+    create_table :user_mentee_applications do |t|
+      t.references :user, null: false, foreign_key: true, index: true
+
+      t.string :city, null: false
+      t.string :state, null: false
+      t.string :country, null: false
+
+      t.text :reason_for_applying, null: false
+      t.string :linkedin_url
```
crespire, line 11: Would it be better here to use a `JSONB` type like `social_urls`? Rails will automatically handle serialization from JSONB to a hash, so you can define things like `social_urls['linkedin'] = "xyz"` and not worry about adding a new column. JSONB is still collection queryable and you can also search for `contains` within the JSONB with [some SQL](https://stackoverflow.com/questions/35737931/rails-postgres-query-with-jsonb-array).

crespire (commented):
Looks great! Just a comment, not a blocker.

db/migrate/20230809025152_create_user_mentee_application.rb
toyhammered: not for this. We plan on handling social urls completely differently in a profile so I want to just keep this simple.

PR #217, opened 2023-08-09
Blog post model

JoshDevHub:
## What's the change?
- Run Rails' builtin installation for ActionText
- Add model and migration for `BlogPost`

Potential surprise: `BlogPost` doesn't have a `content` field! This surprised me at least. Rich text is kept in the new `action_text_rich_texts` table, which holds a `record_id` to the resource it belongs to. This is quite similar to how ActiveStorage works.


toyhammered (approved):
LGTM!

PR #220, opened 2023-08-11
add working controller + new views

toyhammered:
## What's the change?
Added the UserMenteeApplicationController + a few views. None of this is styled.
This should allow multiple people to now help with the work required here (i'll be making some tickets).

### DIFF
--- a/app/controllers/user_mentee_applications_controller.rb
+++ b/app/controllers/user_mentee_applications_controller.rb
@@ -1,0 +1,54 @@
+class UserMenteeApplicationsController < ApplicationController
+  before_action :set_user_mentee_application, only: %i[show]
+
+  def index
+    @user_mentee_applications = policy_scope(UserMenteeApplication)
+  end
+
+  def show
+    authorize @user_mentee_application
+  end
+
+  def new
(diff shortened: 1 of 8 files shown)


afogel (approved):
Looks good. I didn't get around to finishing updating the wireframe mocks in time before you put this up (my cousin is visiting rn), but one thing that could be nice is to organize the form like I did in the mobile mock: https://balsamiq.cloud/se3m7v2/puupn9o/r0AB4

Also, in the form, we probably want to require most of the fields outside of "anything else we should know", right?

toyhammered:
> Looks good. I didn't get around to finishing updating the wireframe mocks in time before you put this up (my cousin is visiting rn), but one thing that could be nice is to organize the form like I did in the mobile mock: https://balsamiq.cloud/se3m7v2/puupn9o/r0AB4
> 
> Also, in the form, we probably want to require most of the fields outside of "anything else we should know", right?

yes. I'll be doing all that in the next PR (organizing, client validations, proper fields as needed). Mainly just wanted to get a couple other stories unblocked so others can work on them.

PR #238, opened 2023-08-15
User mentee application styling

toyhammered:
## What's the change?
Adds styling for the New Page (User Mentee Application).
Properly organizes the data as per the wireframes.

Screenshot 2023-08-14 at 8 10 47 PM
Screenshot 2023-08-14 at 8 11 27 PM

### DIFF
--- a/app/views/user_mentee_applications/new.html.erb
+++ b/app/views/user_mentee_applications/new.html.erb
@@ -7,11 +7,11 @@
     <%= render UserMenteeApplication::SectionComponent.new(header_text: 'Tell us about your journey') do %>
       <div class="flex flex-col mb-5">
         <%= form.label :learned_to_code, 'How did you learn to code?', class: 'font-bold' %>
-        <%= form.text_area :learned_to_code, class: 'textarea textarea-bordered', required: true %>
+        <%= form.text_area :learned_to_code, class: 'textarea textarea-bordered', required: true, data: { controller: 'textarea-autogrow' } %>
       </div>
       <div class="flex flex-col">
         <%= form.label :project_experience, "Describe a project you’ve worked on that you especially loved or are proud of.",  class: 'font-bold' %>
-        <%= form.text_area :project_experience, class: 'textarea textarea-bordered', required: true %>
+        <%= form.text_area :project_experience, class: 'textarea textarea-bordered', required: true, data: { controller: 'textarea-autogrow' } %>
       </div>
     <% end %>
 
@@ -71,7 +71,7 @@
     <%= render UserMenteeApplication::SectionComponent.new(header_text: 'Additional Information') do %>
       <div class="flex flex-col">
         <%= form.label :additional_information, 'Anything else we should know?', class: 'font-bold' %>
-        <%= form.text_area :additional_information, class: 'textarea textarea-bordered' %>
+        <%= form.text_area :additional_information, class: 'textarea textarea-bordered', data: { controller: 'textarea-autogrow' } %>
       </div>
     <% end %>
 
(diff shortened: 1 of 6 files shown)


app/views/user_mentee_applications/new.html.erb
```diff
@@ -1,66 +1,83 @@
...
+  Welcome <%= current_user.first_name %>!
+</h1>
 
-<%= form_with model: @user_mentee_application, local: true, html: { class: 'flex flex-col' } do |form| %>
+<%= form_with model: @user_mentee_application, local: true, html: { class: 'max-w-prose mx-auto' } do |form| %>
+  <div class='flex flex-col ml-5 mr-5'>
+    <%= render UserMenteeApplication::SectionComponent.new(header_text: 'Tell us about your journey') do %>
+      <div class="flex flex-col mb-5">
+        <%= form.label :learned_to_code, 'How did you learn to code?', class: 'font-bold' %>
+        <%= form.text_area :learned_to_code, class: 'form-input border-2 mb-2', required: true %>
```
JoshDevHub, line 10: A lot of these inputs would look a bit better with some Daisy component styles in my opinion. Like [here's the stuff for textareas](https://daisyui.com/components/textarea/). And of course you can use their styles for all the other inputs as well. Can even maybe use the styles with colored borders if you think they look good, as injecting some color on this page would probably spice things up a bit.

app/views/user_mentee_applications/new.html.erb
```diff
@@ -1,66 +1,83 @@
...
+    <% end %>
 
-  <div class="flex flex-col">
-    <%= form.label :city, class: 'font-bold' %>
-    <%= form.text_field :city, class: 'form-input' %>
-  </div>
+    <%= render UserMenteeApplication::SectionComponent.new(header_text: 'Tell us about your interest in the Agency of Learning') do %>
+      <div class="flex flex-col mb-5">
+        <%= form.label :reason_for_applying, 'Why are you here?', class: 'font-bold' %>
+        <%= form.select :reason_for_applying, options_for_select(UserMenteeApplicationsHelper::REASONS_FOR_APPLYING), class: 'form-input mb-2', required: true %>
```
JoshDevHub, line 21: The styles on this `form.select` actually aren't being applied. You can check by inspecting the `<select>` tag's markup and seeing that no classes are there.

This is because of the kinda dumb API they have for this method lol: https://api.rubyonrails.org/v7.0.7/classes/ActionView/Helpers/FormBuilder.html#method-i-select

You basically have to write it like:
```erb
<%# Note the empty options hash among the args %>
form.select :reason_for_applying, options_for_select(reasons), {}, class: 'these classes work', required: true
```

JoshDevHub (changes requested):
I like the layout of it! Just a couple changes I think should be made to have it look a bit more polished.

app/views/user_mentee_applications/new.html.erb
```diff
@@ -1,66 +1,83 @@
...
+  Welcome <%= current_user.first_name %>!
+</h1>
 
-<%= form_with model: @user_mentee_application, local: true, html: { class: 'flex flex-col' } do |form| %>
+<%= form_with model: @user_mentee_application, local: true, html: { class: 'max-w-prose mx-auto' } do |form| %>
+  <div class='flex flex-col ml-5 mr-5'>
+    <%= render UserMenteeApplication::SectionComponent.new(header_text: 'Tell us about your journey') do %>
+      <div class="flex flex-col mb-5">
+        <%= form.label :learned_to_code, 'How did you learn to code?', class: 'font-bold' %>
+        <%= form.text_area :learned_to_code, class: 'textarea textarea-bordered', required: true %>
```
JoshDevHub, line 10: On these textareas, you can use the `textarea-autogrow` Stimulus controller to make it so the box dynamically grows as the user types content that overflows the container. I personally think this looks a bit better than the default scrollbar behavior, but it's also not a huge deal.

You can add this in by adding `data: { controller: 'textarea-autogrow' }` argument to `form.text_area` call.

JoshDevHub (approved):
Awesome. Just one other small, non-blocking comment. Feel free to merge :rocket:

PR #239, opened 2023-08-16
Featured blog

### LINKED ISSUE #230: "Featured" Blogs

JoshDevHub:
## What's the change?
- Add model for Featured Blog Posts
- Add `ranked-model` gem to help manage the ordering of featured blog posts
- Add ability for an Admin user to click a button to "feature" a blog post while viewing that post
- Add pundit policy for authorizing user actions for featured posts
- Add publicly viewable link to Featured Blog in the welcome page

Migrate down to cleanup

A potential surprise is that there's currently no frontend flow for admins to change the ordering of featured posts. In my mind, this would be a bit complicated to do well. I think it's probably fine for now to have admins interface with this through the production console?

#### Feature Post Button
image

#### Remove Feature Button
Selection_246

Closes #230 


JoshDevHub:
Marking this as ready for review. I had maybe one question develop in my mind while writing this: is `FeaturedBlogPost` a good name for this new model? Just feels weird to say something like `featured_post.blog_post`. Maybe something like `FeaturedTag` or `FeaturedBlogTag` or something would be better? Or maybe it's just fine as is. I'm interested in thoughts regarding that.

toyhammered:
> Marking this as ready for review. I had maybe one question develop in my mind while writing this: is `FeaturedBlogPost` a good name for this new model? Just feels weird to say something like `featured_post.blog_post`. Maybe something like `FeaturedTag` or `FeaturedBlogTag` or something would be better? Or maybe it's just fine as is. I'm interested in thoughts regarding that.

I mean we can always make it read like `featured_post.post` or something. I think FeaturedBlogPost is fine for now. It is intuitive what it is, and it will be super easy to change later on (and doesn't take up any generic namespace)

app/controllers/blog_posts_controller.rb
```diff
@@ -1,6 +1,7 @@
 class BlogPostsController < ApplicationController
   def show
     @blog_post = authorize BlogPost.find(params[:id])
+    @featured_blog_post = FeaturedBlogPost.find_or_initialize_by(blog_post: @blog_post)
```
toyhammered, line 4: how come we are doing a `find_or_initialize_by`, shouldn't this exist already?

toyhammered (approved):
Love it! Keep on crushing it!

app/controllers/blog_posts_controller.rb
JoshDevHub: So the post that's being shown may not be "featured", in which case a `FeaturedBlogPost` will be initialized here. If it is featured, then that will be looked up.

This basically sets up the data so that I can render buttons for admin users to "feature" or "remove feature" on the post. Pundit policies want to have an object, which is why it's nice to initialize one if the blog hasn't been featured yet. I can still say `if policy(@featured_blog_post).create?` and have things work nicely. Hopefully that makes sense.

toyhammered: hmmm still seems kind of strange. Like can't we just force the policy check to use a specific policy (i.e: the featured one for that blog post). As for the check, can't we just do like an `includes(:featured_blog_post)` and then have a method like if `blog_post.featured?` or something?

JoshDevHub: Hmm yeah that `featured?` method idea sounds better. I'll cook up a refactor real quick.

JoshDevHub: Alright managed to find enough time tonight to get some changes in. I ended up still being able to call the policies normally anyway. The `#featured?` method is good though and in hindsight, probably something I should've considered 😵‍💫

toyhammered: it's all good! We are all here to learn and glad I could teach you something new.

PR #243, opened 2023-08-18
fix: Profile slug - add validations and move FriendlyID errors

crespire:
## What's the change?
Update profile to make sure slug is unique (we want to give the user an error rather than silently fix it via FriendlyID).

Also make sure we move the FriendlyID error to the slug field if a user enters a reserved word (built in with FriendlyID)
Profile edit/update flow.


JoshDevHub (approved):
LGTM!

PR #244, opened 2023-08-18
Issue#203 check in button

RocDavenport:
Added a check-in button to standup_meeting_group.

When a logged in user visits the standup page, they will now have the ability to check-in.

## Demo / Screenshots
Screen Shot 2023-08-22 at 1 49 41 PM
Screen Shot 2023-08-22 at 1 59 56 PM

https://github.com/agency-of-learning/PairApp/compare/issue%23203_check_in_button

bin/check was ran and completed.

- [n ] Did you add appropriate automated tests?
- [n ] Did you consider risks around security, performance, etc.?
- [n ] Have you thought of misfiring code? e.g. too many loops, n+1, or how to handle nils?
- [n ] Any validations to data needed?

- [y ] Is everything elevated to the highest level? (e.g., can logic be moved out of view into controllers, or out of controllers into models?)

- [ n] Did you leave helpful inline PR comments for the reviewer(s)?
- [n ] Did you include instructions for how to test in the ticket?
- [ n] Did you add any relevant tags and decide if this PR is a Draft vs. Ready for Review?
- [n ] Do you have a plan for how this change should be rolled back? (e.g., how do you deal with a migration rollback? Is your feature released behind a feature flag?)


spec/components/standup_meeting_group/check_in_button.html.erb
```diff
@@ -0,0 +1 @@
+<%= link_to 'Check In', standup_meeting_group_standup_meetings_path(@standup_meeting_group), class: "btn btn-primary"%>
```
JoshDevHub, line 1: Both of these `spec/` files don't really need to exist if you're not testing anything here.

app/components/standup_meeting_group/check_in_button_component.rb
```diff
@@ -0,0 +1,6 @@
+class StandupMeetingGroup::CheckInButtonComponent < ViewComponent::Base
+  def initialize(standup_meeting_group)
+    super
```
JoshDevHub, line 3: You don't have to call `super` in ViewComponent constructors.

JoshDevHub (changes requested):
Couple small changes that should be made.

But also: I might want clarity on is what the intended behavior of this button is. Should it take the user to today's standup meeting for that group? This is of course how you have it working currently. Or should it take the user to the form to fill in today's standup meeting? This is how I was thinking it should work, as it's how the other "Check-In" buttons work. But I'm also not 100% sure.

Going to tag @afogel @toyhammered to get their thoughts on it since they did the wireframes for this feature.

app/components/standup_meeting_group/check_in_button_component.rb
RocDavenport: Again, I agree. Thank you for this Josh. I really was trying to understand the error warning that I was receiving when running the checks. Again, I will make the appropriate changes.

spec/components/standup_meeting_group/check_in_button.html.erb
RocDavenport: @JoshDevHub  "If that sounds worth pursuing to you, I'd be glad to pair on it and help out." Uhhh. Yes please.

afogel:
> Couple small changes that should be made.
> 
> But also: I might want clarity on is what the intended behavior of this button is. Should it take the user to today's standup meeting for that group? This is of course how you have it working currently. Or should it take the user to the form to fill in today's standup meeting? This is how I was thinking it should work, as it's how the other "Check-In" buttons work. But I'm also not 100% sure.
> 
> Going to tag @afogel @toyhammered to get their thoughts on it since they did the wireframes for this feature.
> 
> Edit: From looking at the Issue requirements #244 -- It seems Gustavo thinks it should take the user directly to the form. Also from a little poking around the code, there may be a way to refactor some of the stuff around this ticket. There's already [this `CheckInCtaComponent`](https://github.com/agency-of-learning/PairApp/blob/main/app/components/standup_meeting/check_in_cta_component.html.erb) with some of the same info and buttons we're wanting to display for the user on the `StandupMeetingGroup::JoinedCardComponent`s, but in a different layout. There's probably some way to have this component fit in, or maybe have a more generic component that fits both layouts. If that sounds worth pursuing to you, I'd be glad to pair on it and help out.

I'd echo what Josh said -- the expectation is that clicking the checkin button takes you directly to the form for `Date.current` and clicking the Skip button should be a form submission that skips for `Date.current`. Awesome that you're working on this -- I just put up a pr ( #249 ) that improves the UX of the index page and I was about to turn to making the checkin/skip buttons work when I saw you'd put up a PR already!

JoshDevHub:
Me and Roc paired on new changes around this. The CTA and button group have been refactored to fit being in the `StandupGroup#index` and `StandupGroup#show` without having completely different components to handle those cases. Flagging others for review since I participated in these changes.

toyhammered (approved):
This looks great to me! Just left 1 question.

app/components/standup_meeting/check_in_action_component.html.erb
```diff
@@ -0,0 +1,13 @@
+<% if standup_meeting.completed? %>
+  <%= link_to "Edit Response", 
+              edit_standup_meeting_group_standup_meeting_path(standup_meeting_group, standup_meeting), 
+              class: 'btn btn-primary' %>
+<% elsif standup_meeting.persisted? %>
+  <%= link_to "Check in",
+              edit_standup_meeting_group_standup_meeting_path(standup_meeting_group, standup_meeting), 
+              class: 'btn btn-primary' %>
+<% else %>
+  <%= button_to "Check in",
```
toyhammered, line 10: why the same text here?

JoshDevHub: So for one case, the daily standup has already been created (ie via background worker) and this needs to link to the edit page for that existing standup. The other case is that it doesn't exist yet and needs to be created.

These were previously both called 'Check in', and I didn't immediately see a problem with it. I could be down to change them if you have good names for them though.

PR #247, opened 2023-08-19
Friendly Routing for Blog Posts

### LINKED ISSUE #231: [Blog Epic] Friendly Routing to Blog Posts
Users want to be able to link to a blog post by a human reader friendly slug instead of the blog post's id. A good candidate to slug the blog posts by would be their title.

JoshDevHub:
## What's the change?
- Add friendly id slugging for blog posts; posts are slugged by their title
- Adjust authorization permissions so non-logged in users can visit featured blog posts

Visiting blog posts

A potential surprise is `Null::User`. Users who don't have an account are now able to visit certain pages on the site. The null user allows Pundit to not have to check `present?` constantly to handle a user not being logged in. Its API is currently light, but it could grow to handle more cases as needed.

Closes #231 


app/models/blog_post.rb
```diff
@@ -34,7 +37,22 @@ class BlogPost < ApplicationRecord
...
   def featured?
     featured_blog_post.present?
   end
+
+  def should_generate_new_friendly_id?
+    title_changed?
+  end
+
+  def initialize_feature
+    featured_blog_post || build_featured_blog_post
```
JoshDevHub, line 56: This method might seem a little strange but I needed it because of some quirks with `has_one`. If you use `build_association` on an object that already has an associated object, the associated object will be destroyed (like actually deleted from the database). This method allows getting the associated object if it exists or building it if not, which avoids that side effect.

GALTdea: Just curious, Is there a particular reason why we're making this a has_one featured_blog_post?

GALTdea: Oh, I see each blog can only be featured once

afogel: I find this `initialize_feature` naming a bit confusing -- is it meant to create a feature or access a feature?
If the latter, it might be better to rename to `featured_post`, since it's returning the blog post. Or alternatively, perhaps overwrite the `attr_reader` method for `featured_blog_post` to access the associated blog post if exists, else build a new one.

As I write this out, I have some spidey-sense that's ringing re: feature envy within the BlogPost class -- seems like it may know a bit too much about this featured blog post model

afogel (commented):
Not blocking since I don't want perfect to be the enemy of good enough, but I do have some concerns I raised around how featured blog posts are handled. Would be curious about @toyhammered thoughts, too.

app/policies/application_policy.rb
```diff
@@ -4,7 +4,7 @@ class ApplicationPolicy
   attr_reader :user, :record
 
   def initialize(user, record)
-    @user = user
+    @user = user || Null::User.new
```
toyhammered, line 7: I am assuming the use case for the `Null::User.new` is for users not signed in. and pages that include authorization (but allow non signed in users). For this case I would just have a `return false if user.blank?` on these specific areas. I don't like what this object is doing at all (and it isn't entirely correct either)

app/views/blog_posts/show.html.erb
```diff
@@ -6,9 +6,9 @@
         data: { turbo_method: :delete },
         class: "btn btn-error btn-sm"
       %>
-    <% elsif policy(@blog_post.build_featured_blog_post).create? %>
+    <% elsif policy(@blog_post.initialize_feature).create? %>
```
toyhammered, line 9: I am pretty sure we had a similar issue somewhere in the past with the `policy` view helper not being allowed to route to a specific policy (it has to be what the record inference is). Is this the reason you do this round about way?

If so, I think at this point we might as well add our own helper that somewhat overrides this to offer the correct functionality.

toyhammered (changes requested):

app/views/blog_posts/show.html.erb
toyhammered: I will also say this is one of the few use cases that we can have a non crud policy check because this is super specific to what it does.

JoshDevHub: Yeah this was purely with the intent to keep the policies CRUD. If you're good with a `#feature?` policy method on blog posts, I could absolutely do it that way.

toyhammered (approved):

PR #249, opened 2023-08-20
allow join/leave from groups#index page

afogel:
## What's the change?

Currently, the UX on the groups#index page has some behavior that has jarring UX; namely, the list of possible options disappears, rather than making it clear what are the actions. This inconsistency of making clear how to join/leave from the same spot breaks with UX conventions, and makes it difficult for a user to identify where they can leave a standup meeting group (they have to know, and/or search, for it on the group#show page). 

This PR introduces a symmetric flow for leaving a group from the same place where users join, improving the usability of the workflow.

Additionally, there are some minor change wrt styling/polish that are made here (cards should not be centered by default, add box shadow to improve readability).

Joining/Leaving a standup meeting group.

## Demo / Screenshots
[Edurrhaphy_Recording_2023_08_20-10_55.webm](https://github.com/agency-of-learning/PairApp/assets/2447409/80b69f99-7a4f-497a-bc48-07bce454536b)

^ For this, I realized that there may be a need to add a unique index on the db level for standup meeting groups; currently, there is nothing in the system that stops a user from joining the same group multiple times. I think, however, that this is out of scope for this PR and should be addressed in a different issue.

- [X] Can (and should) any models be extracted?
- [X] Did you include instructions for how to test in the ticket?

### DIFF
--- a/app/views/standup_meeting_groups/index.html.erb
+++ b/app/views/standup_meeting_groups/index.html.erb
@@ -6,7 +6,7 @@
 
 <div class="flex flex-col justify-center">
   <h2 class="font-bold text-3xl pl-5 text-center md:text-left">My Standups</h2>
-  <div id="my-standup-meeting-groups" class=" flex justify-center flex flex-wrap gap-5 pl-5 pt-5 pb-5">
+  <div id="my-standup-meeting-groups" class="flex flex-wrap gap-5 pl-5 pt-5 pb-5">
     <%= render StandupMeetingGroup::JoinedCardComponent.with_collection(@my_standup_meeting_groups, current_user:) %>
   </div>
 </div>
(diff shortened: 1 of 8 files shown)


app/views/standup_meeting_groups/index.html.erb
```diff
@@ -6,14 +6,14 @@
...
   <h2 class="font-bold text-3xl pl-5 text-center md:text-left">My Standups</h2>
-  <div id="my-standup-meeting-groups" class=" flex justify-center flex flex-wrap gap-5 pl-5 pt-5 pb-5">
+  <div id="my-standup-meeting-groups" class="flex flex-wrap gap-5 pl-5 pt-5 pb-5">
     <%= render StandupMeetingGroup::JoinedCardComponent.with_collection(@my_standup_meeting_groups, current_user:) %>
   </div>
 </div>
 
 <div class="flex flex-col">
  <h2 class="font-bold text-3xl pl-5 text-center md:text-left">Find a Standup</h2>
   <ul id="find-standup-meeting-groups" class="flex flex-col pl-5 pt-5 pb-5">
```
JoshDevHub, line 16: Not a blocker, and not even something your changes are responsible for, but before merge, I think it'd be cool to adjust the padding here so the Joinable cards look a bit better on mobile.

Selection_247

JoshDevHub (approved):
LGTM. Definitely don't have to fix my comment and you can feel free to merge. Just a small thing I noticed while looking at this.

toyhammered:
This looks great!

PR #254, opened 2023-08-22
Developing a style guide for brand cohesion across application

afogel:
## What's the change?
I noticed when reviewing the latest changes that there are a number of customized styles that are thrown in. Though perhaps okay in a pinch, they aren't in the existing theme and provide a poor example to future developers of how to introduce user-facing styling that remains consistent with existing styles in the app. 

Since we began to leverage AOL's brand colors for the landing page, I propose a more standardarized approach:
Updating the tailwind theme to override the defaults, using AOL green as the default primary values, AOL grays as the neutral values, and introducing a muted blue (Picton Blue) as a secondary theme.

In addition, I have done an audit of the application's styling to introduce more consistency in the way we use color for flat cards, table headers, and buttons.

For the user, there are none. For us as developers, we should be reviewing CSS carefully in our code reviews as well. Design smells could be things like:
- using a custom color and the `[]` notation rather than one of the existing theme values, e.g. `bg-[#38d68e]`
- using a base color rather than a theme color, e.g. `text-white` instead of `text-primary-content`

## Demo / Screenshots
Screen Shot 2023-08-22 at 3 42 27 PM
Screen Shot 2023-08-22 at 3 42 20 PM
Screen Shot 2023-08-22 at 3 42 13 PM
Screen Shot 2023-08-22 at 3 42 06 PM
Screen Shot 2023-08-22 at 3 41 59 PM
Screen Shot 2023-08-22 at 3 41 51 PM
Screen Shot 2023-08-22 at 3 42 46 PM


JoshDevHub:
I think it looks nice! Kind of just have one question: do you think the success flash messages still look okay with a green that's close to the "brand" color?

toyhammered:
> I think it looks nice! Kind of just have one question: do you think the success flash messages still look okay with a green that's close to the "brand" color?

I had the exact same thoughts funnily enough. Maybe we should do a darker green for success messages?

toyhammered (commented):
Can be done in a separate PR but do we also want to get some fonts included also?

afogel:
@toyhammered I think it shouldn't be too bad to include fonts here. What did you have in mind?
As for the success, yeah, I'll play with it to make it look more distinct.
Thanks for the quick review @toyhammered and @JoshDevHub !

toyhammered:
> @toyhammered I think it shouldn't be too bad to include fonts here. What did you have in mind? As for the success, yeah, I'll play with it to make it look more distinct. Thanks for the quick review @toyhammered and @JoshDevHub !

I like roboto :D

elhalvers (commented):
Great call on creating standard theme colors for better consistency throughout the app!

https://www.color-hex.com/color/38dc8e : )

crespire (approved):
Great idea!

dpaola2:
Great idea @afogel JFDI 💪

Issue #225, opened 2023-08-11
[UserMenteeApplication] Allow for admin to update the status of an application


toyhammered:
There is a partial proposal here: https://www.notion.so/toyhammered/Application-Submission-Design-Doc-7a45fd8e25ce4f448e9b47bc00a142b5?pvs=4#be8855172d204e2a9e46a10146042e8f

On the index page, admin should be allowed to progress someone to a new state and leave notes. This will also be recorded in our db so we know which admin made what decision.

This same functionality should be on the show page also.

PR #263, opened 2023-08-26
WIP:#225 add admin status to mentee application

GALTdea:
## What's the change?
I am making this a draft PR because there's a few considerations needed for the "Nexting" "Rejecting" flow. 
I'll continue to work on the Note aspect of the feature. 

It gives admins the ability to move mentee applications to the next stage aswell as rejecting"
## Demo / Screenshots

https://github.com/agency-of-learning/PairApp/assets/16809030/6563f460-d623-4f9a-89e7-5018d4c1c583

https://github.com/agency-of-learning/PairApp/assets/16809030/01b370dd-d67c-446b-a4d5-43776199ad67

mentee application show  page
Screen Shot 2023-08-30 at 2 08 41 PM

https://github.com/agency-of-learning/PairApp/assets/16809030/4083e006-064a-43ac-a515-29c705c1815e

#225 

- [ ] Did you add appropriate automated tests?
- [ ] Did you consider risks around security, performance, etc.?
- [ ] Any validations to data needed? Maybee? 


GALTdea:
Questions:
1. From yesterday's meeting, it seems like applications will need at least two members approving the status change. Do we want to integrate that feature in the flow? 

2. Do we want to allow for rejected applications to be restored? 

3. Do we want to allow for Accepted applications to be cancelled? 

4. Do we really want to be able to promote or reject applications from the index page, my proposal is that we do it from the show page so we minimize the risk of mistakenly clicking on the wrong button
@toyhammered @dpaola2

app/models/mentee_application_state.rb
```diff
@@ -0,0 +1,42 @@
...
+    pending: 0,
+    stage_one: 1,
+    stage_two: 2,
+    stage_three: 3,
+    stage_four: 4,
+    accepted: 5,
+    rejected: 6
+  }.freeze
+
+  enum status: STATUSES
```
JoshDevHub, line 34: You may already know this, but you don't have to write it this way to be able to refer to the enum hash
```rb
enum status: {
  pending: 0,
  stage_one: 1,
  # etc.
}
  
MenteeApplication.statuses
#=> { "pending" => 0, "stage_one" => 1 } # etc.

# the returned hash also has 'indifferent access', which means you can key in with strings or symbols
MenteeApplication.statuses["pending"] #=> 0
MenteeApplication.statuses[:pending] #=> 0
```
Just some information if you didn't know.

app/views/user_mentee_applications/_table_row.html.erb
```diff
@@ -8,16 +8,17 @@
...
 
   <%= render TableCellComponent.new(container_link: mentee_application) do %>
-    <%= mentee_application.country %>
+    <%#= mentee_application.notes %>
   <% end %>
-
+  
   <%= render TableCellComponent.new(container_link: mentee_application) do %>
-    <%#= mentee_application.notes %>
+    <%= mentee_application.current_status %>
```
JoshDevHub, line 15: Ultra small thing, but you can call `humanize` on this to make it look a bit more natural on the output. This being an admin-facing feature, I don't think it's as important to eliminate all code-ish looking stuff though.
```rb
mentee_application.current_status #=> "stage_one"
mentee_application.current_status.humanize #=> "Stage one"
```

app/controllers/user_mentee_applications_controller.rb
```diff
@@ -24,6 +24,7 @@ def create
 
     @user_mentee_application = UserMenteeApplication.new(user_mentee_application_params)
     @user_mentee_application.user = current_user
+    @user_mentee_application.mentee_application_states.build(status: :pending)
```
JoshDevHub, line 27: Would there ever be a situation where a `mentee_application` exists without having any `mentee_application_state`? I only ask because if this situation is impossible, then I might think it's a good idea to create the associated default `mentee_application_state` in an `after_create` callback. Not sure on this though, just something I'm curious about.

JoshDevHub (commented):
Left some comments about the code. I'll leave deeper examination to Daniel since he's much more familiar with this feature and the domain design of it. It's looking pretty good.

toyhammered:
> Questions:
> 
> 1. From yesterday's meeting, it seems like applications will need at least two members approving the status change. Do we want to integrate that feature in the flow?
> 2. Do we want to allow for rejected applications to be restored?
> 3. Do we want to allow for Accepted applications to be cancelled?
> 4. Do we really want to be able to promote or reject applications from the index page, my proposal is that we do it from the show page so we minimize the risk of mistakenly clicking on the wrong button
>    @toyhammered @dpaola2

To answer all your questions:
1. we are not going to be doing that for this first phase (it adds quite a few complications not worth the trouble).
2. no
3. hmmmm... I'd say leave that out for now but I do like this idea of "we accepted someone but they were not a good fit".
4. Yes. What we should do though is have a popup that says something like "do you really want to move this person to hte next stage", etc... Very similar how you would do so with a delete action. This will also allow us to test both flows out and if we notice that we are utilizing one more than the other we can always refactor.

app/components/user_mentee_application/action_button_builder_component.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= render_button %>
+
+<%= render_reject_button %>
```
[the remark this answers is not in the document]
GALTdea: oh, nice! didn't know you could do that. Thank you

app/controllers/user_mentee_applications_controller.rb
GALTdea: I was trying not to use a callback but this approach is giving me test issues. I'll try the callback.

app/models/mentee_application_state.rb
GALTdea: hm, I figured it would be good to have an independent constant with different statuses, but now I'm wondering if it's just repeating code?

app/controllers/user_mentee_applications_controller.rb
GALTdea: Ok, after trying both, I think I lean more towards the callback approach.

app/models/mentee_application_state.rb
toyhammered: what Josh said is correct and that is the general approach.

GALTdea:
Ok, 
I added a small navbar to the user_mentee_application show page. 
Still needs styling but functionality wise it is working fine. 
@toyhammered 
**Notes:** 
For the note feature, I believe we need to make some design decisions because the action buttons as I have right now may not work. 

The issue is that the link only allows me to update the application status but won’t send the note’s data.

The lowest hanging fruit is to add form instead of links. 
I can use checkboxes for the reject and “next stage” fields
And add a text_area for the notes. 

And that way the form will get submitted with  all the data needed.

toyhammered:
> Ok, I added a small navbar to the user_mentee_application show page. Still needs styling but functionality wise it is working fine. @toyhammered **Notes:** For the note feature, I believe we need to make some design decisions because the action buttons as I have right now may not work.
> 
> The issue is that the link only allows me to update the application status but won’t send the note’s data.
> 
> The lowest hanging fruit is to add form instead of links. I can use checkboxes for the reject and “next stage” fields And add a text_area for the notes.
> 
> And that way the form will get submitted with all the data needed.

Maybe for the initial could we just have the `notes` be a separate form. (i.e: textbox with some OK button next to it)

GALTdea:
> > Ok, I added a small navbar to the user_mentee_application show page. Still needs styling but functionality wise it is working fine. @toyhammered **Notes:** For the note feature, I believe we need to make some design decisions because the action buttons as I have right now may not work.
> > The issue is that the link only allows me to update the application status but won’t send the note’s data.
> > The lowest hanging fruit is to add form instead of links. I can use checkboxes for the reject and “next stage” fields And add a text_area for the notes.
> > And that way the form will get submitted with all the data needed.
> 
> Maybe for the initial could we just have the `notes` be a separate form. (i.e: textbox with some OK button next to it). If this would work, feel free to do this functionality in a different PR.

hm, do you want me to add the comments into it's own user_mentee_state?
or I can always add the note to the current_application_state but that can present different issues

app/models/user_mentee_application.rb
```diff
@@ -28,9 +29,60 @@
...
+  end
+
+  def can_promote?
+    current_index = MenteeApplicationState::STATUSES.keys.index(last_status)
+    max_index = MenteeApplicationState::STATUSES.keys.length - 2
+    current_index < max_index && last_status != :rejected
+  end
+
+  def rejected?
+    last_status == :rejected
```
JoshDevHub, line 72: This can be delegated to `current_state`. Same with the `accepted?` method. Let me know if you want me to explain more about how this works.
```rb
delegate :rejected?, to: :current_state
```

app/models/user_mentee_application.rb
```diff
@@ -28,9 +29,60 @@
...
+
+  def current_state
+    mentee_application_states.last
+  end
+
+  def next_status
+    current_status_index = MenteeApplicationState::STATUSES.keys.index(current_status.to_sym)
+    next_status_index = current_status_index.to_i + 1
+
+    MenteeApplicationState::STATUSES.keys.fetch(next_status_index)
```
JoshDevHub, line 53: Outside of the `STATUSES` thing we've already talked about, I think this method is good! One small problem I have is though is the implicit understanding that this method doesn't work and doesn't make sense for `accepted` and `rejected` statuses. It doesn't make sense for the `accepted` status because the calculated next status will be `rejected`, and if you try this method with a `rejected` status, it will error on the `fetch` going out of bounds.

I think it'd be better to have an explicit exception for these two. Maybe have an early return if the current status is accepted or rejected?

JoshDevHub (changes requested):
This is looking really good. Just a few more requested changes. Let me know if you have questions about anything. Also glad to meet/pair anytime you want to talk something over.

app/components/user_mentee_application/action_button_builder_component.rb
```diff
@@ -0,0 +1,29 @@
...
+
+class UserMenteeApplication::ActionButtonBuilderComponent < ViewComponent::Base
+  attr_reader :mentee_application
+
+  def initialize(mentee_application:)
+    @mentee_application = mentee_application
+  end
+
+  def render_button
+    return unless @mentee_application.can_promote?
```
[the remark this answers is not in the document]
GALTdea: aw, yes.  This was a more involved method at one point, but you are right in that it's not needed any more. thank you for catching it!

app/models/user_mentee_application.rb
```diff
@@ -28,9 +29,60 @@
...
+  end
+
+  def reject_application(current_user)
+    mentee_application_states.build(status: :rejected, status_changed_by_id: current_user.id).save
+  end
+
+  def can_promote?
+    current_index = MenteeApplicationState::STATUSES.keys.index(last_status)
+    max_index = MenteeApplicationState::STATUSES.keys.length - 2
+    current_index < max_index && last_status != :rejected
```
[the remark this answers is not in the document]
GALTdea: very great point!

db/migrate/20230823001215_create_mentee_application_states.rb
```diff
@@ -0,0 +1,12 @@
+class CreateMenteeApplicationStates < ActiveRecord::Migration[7.0]
+  def change
+    create_table :mentee_application_states do |t|
+      t.references :user_mentee_application, null: false, foreign_key: true
+      t.integer :status, null: false, default: 0
+      t.text :note
+      t.integer :status_changed_by_id
```
toyhammered, line 7: this should be some type of t.references with like `foreign_key: { to_table: :users }` or something like that.

db/migrate/20230821233937_add_status_to_user_mentee_applications.rb
```diff
@@ -0,0 +1,5 @@
+class AddStatusToUserMenteeApplications < ActiveRecord::Migration[7.0]
+  def change
+    add_column :user_mentee_applications, :status, :integer
+  end
+end
```
toyhammered, line 5: We don't need this. There shouldn't be a `status` on the application directly.

app/views/user_mentee_applications/show.html.erb
```diff
@@ -58,4 +68,12 @@
...
\ No newline at end of file
+</div>
+
+
+
+
+
+
+
+
```
toyhammered, line 79: dont need these new lines?

app/models/mentee_application_state.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  enum status: {
+    pending: 0,
+    stage_one: 1,
+    stage_two: 2,
+    stage_three: 3,
+    stage_four: 4,
+    accepted: 5,
+    rejected: 6
+  }
```
toyhammered, line 34: Not in this PR, but we will need to figure out the actual flow. For this initial PR it might be worth changing from `pending` to `application_received` as the starting point.

toyhammered (changes requested):
Looks great. Couple small things to fix up!

app/models/user_mentee_application.rb
```diff
@@ -28,9 +29,50 @@
...
+
+  def next_status
+    return if accepted? || rejected?
+    current_status_index = MenteeApplicationState.statuses.keys.index(current_status.to_sym)
+    next_status_index = current_status_index.to_i + 1
+
+    MenteeApplicationState.statuses.keys.fetch(next_status_index)
+  end
+
+  def promote_application(user)
```
afogel, line 61: I would add a `!` to the method call, since it's a destructive action.

app/models/user_mentee_application.rb
```diff
@@ -28,9 +29,50 @@
...
     :available_hours_per_week, presence: true
+
+  after_create :create_initial_application_state
+
+  delegate :rejected?, to: :current_state
+
+  delegate :accepted?, to: :current_state
+
+  def current_status
+    mentee_application_states.last&.status&.to_sym
```
afogel, line 46: why are we using the safe navigation operator here? Shouldn't there always be an associated state with an application?

app/models/user_mentee_application.rb
```diff
@@ -28,9 +29,50 @@
...
+  end
+
+  def current_state
```
afogel, line 49: seems like this should be `private`, too.

app/models/user_mentee_application.rb
```diff
@@ -28,9 +29,50 @@
...
+    mentee_application_states.last
+  end
+
+  def next_status
```
afogel, line 53: seems like this should be private.

afogel (changes requested):
seems like this is a complex state machine distributed across two models. I think we should probably expose as little as possible about how the model transitions from one state to the next in its API.

afogel (changes requested):
At a high level, I'd like to say that I think that most of the state transition logic should be moved to the `MenteeApplicationState` model, including the `next_status` method.

It seems reasonable that a user_mentee_application should `has_one :current_state` (which could be defined elegantly using a scope, which also enables preloading).

In this way, most of the rest of messages relating to state progression knowledge should be delegated to the model responsible for knowing about the state machine and its various transitions.

This would simplify the logic in the UserMenteeApplication to something like (i'm just winging the `raise`, here):
```ruby
def promote!(user:)
    raise "Can't promote #{self}" unless can_promote?
    status = MenteeApplicationState.next_status(current_state)
    mentee_application_states.build(status:, status_changed_id: user.id).save!
end

def reject!(user:)
    raise "Can't reject #{self}" unless can_promote?
    status = :rejected
    mentee_application_states.build(status:, status_changed_id: user.id).save!
end
```

app/models/user_mentee_application.rb
afogel: The more I think about it, the more this method seems like presentational/view logic that is mostly duplicating the `current_state` logic. Can we consolidate?

app/models/mentee_application_state.rb
GALTdea: Good point, I think we definitely need to brainstorm the slow

app/models/user_mentee_application.rb
GALTdea: For your first question: yes, good point, I iterated on this method a couple of times but eventually fixed this issue outside  outside of this method. don't think i need it anymore.

for the second: hmm, i'm not sure I get the question. can you elaborate?  This method checks the status of an user_mentee_application record. I use for view logic and in other methods, the state returns the whole mentee_application_state record.

GALTdea:
> At a high level, it seems like the responsibility of most of the state transition logic should be moved to the `MenteeApplicationState` model, including the `next_status` method.
> 

good question: 
- while the status does belong to the mentee_application_state, i think the transitioning logic is done at the user_state_application level/scope. 


> It seems reasonable that a user_mentee_application should `has_one :current_state` (which could be defined elegantly using a scope, e.g., https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_one-label-Scopes).
- UserStatesApplicatin has_many mentee_states. When an application is "transitioned" we query for the last mentee_states and look at it's state and based on that we allow for the next step. I supposed we can do it, but we would need the user_mentee_application Id and then do the same query. 

> In this way, most of the rest of messages relating to state progression knowledge should be delegated to the model responsible for knowing about the state machine and its various transitions.
> 
> This would simplify the logic in the UserMenteeApplication to something like (i'm just winging the `raise`, here):
> 
> ```ruby
> def promote!(user:)
>     raise "Can't promote #{self}" unless can_promote?
>     status = MenteeApplicationState.next_status(current_state)
>     mentee_application_states.build(status:, status_changed_id: user.id).save!
> end
 This would work, but we'd need to passin the mentee_application_state id to figure out it's state

> def reject!(user:)
>     raise "Can't reject #{self}" unless can_promote?
>     status = :rejected
>     mentee_application_states.build(status:, status_changed_id: user.id).save!
> end
> ```

app/models/user_mentee_application.rb
afogel: so, for the second: `current_state` and `current_status` are roughly the same, so I think we should choose one. Given that it is querying the `status` of the `MenteeApplicationState`, I'd recommend `current_status`.

Re: the view logic aspect, I'm wary of adding methods that support view logic to models to combat model bloat. Instead, if there need to be transformations of data to a more human-readable format, I tend to reach for presenters, decorators, or view helpers to define those methods.

Does that make sense/seem reasonable?

db/migrate/20230821233937_add_status_to_user_mentee_applications.rb
GALTdea: oh yeah 😅 forgot to delete this migration

app/models/user_mentee_application.rb
GALTdea: you man to the promote_application method? It creates a new record

app/models/user_mentee_application.rb
GALTdea: I think so, but why don't we pair on this? I have time tomorrow am? 9am pacific? or let me know what time works better for you

JoshDevHub:
To leave my opinion on the API design:

The way Ariel has it is my preference in a vacuum. It's better OO design to have the state understand most of the implementation details and have the mentee application delegate that work to it.

I will say that for me, not all tight coupling has the same cost, which is why I didn't bring it up in my previous reviews. I think with this situation it comes with very low cost. We can pretty safely predict that other models won't be associated with a `MenteeApplicationState`, meaning the locus of specific pieces of the business logic is less meaningful to me. So it's probably something I could go either way on.

toyhammered:
> To leave my opinion on the API design:
> 
> The way Ariel has it is my preference in a vacuum. It's better OO design to have the state understand most of the implementation details and have the mentee application delegate that work to it. And I think it's nice to have consistent application of good OO patterns in the app.
> 
> I will say that for me, not all tight coupling has the same cost, which is why I didn't bring it up in my previous reviews. I think with this situation it comes with very low cost. We can pretty safely predict that other models won't be associated with a `MenteeApplicationState`, meaning the locus of specific pieces of the business logic is less meaningful in a practical sense.
> 
> So yeah I could be on board with the change. But if it disrupts Gustavo's velocity on cranking out these features, I'm less on board for the aforementioned cost reasons.

Love it! I feel no need to step in here and will let you guys handle this yourselves. I'd like to get this feature out sooner rather than later, and I am more than happy if we setup another ticket to refactor this into whatever state we think would be best. 

Let me know if you need me to actually step in or give a final verdict.

afogel:
Thanks for weighing in, Josh and Daniel. I feel strongly that these objections should be addressed. I think that the issues I raise are more of an architectural approach and, as such, I wouldn't categorize them as nits. Additionally, given that the core functionality already exists in the PR, the fix doesn't take long and they provide good examples to future readers of the code. 

Because, however, there seems to be a concern about velocity, and given the timezone difference, I won't be able to pair on this, I have whipped up a PR that addresses these concerns: https://github.com/agency-of-learning/PairApp/pull/280

After my request for better encapsulation is addressed, let's ship this thing.

afogel:
Thanks for weighing in, Josh and Daniel. I still feel that these concerns should be addressed. I think that the issues I raise are related to an architectural approach and, as such, I wouldn't categorize them as nits. Given that the core functionality already exists in the PR, the fix doesn't take long to implement and it will provide better examples to future junior developers/readers of the codebase that we plan to onboard.

Because, however, there seems to be a concern about velocity, and with the timezone difference, I won't be able to pair on this before the end of my workday, I have whipped up a PR that addresses these concerns: https://github.com/agency-of-learning/PairApp/pull/280

@GALTdea please let me know if you want to walk through this asynchronously? I think we can finish this quickly without hampering your velocity too much and ship the feature.
Thanks, all!

GALTdea:
> 

@afogel I agree and that would be great! let me make sure i answer any other feedback and I'll get back to you. Meanwhile can we async in the design doc? that would give a more detailed  Idea of what you're thinking

spec/models/mentee_application_state_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_...  (status_changed_id => users.id)
+#  fk_rails_...  (user_mentee_application_id => user_mentee_applications.id)
+#
+require 'rails_helper'
+
+RSpec.describe MenteeApplicationState do
+  pending "add some examples to (or delete) #{__FILE__}"
```
afogel, line 26: this needs tests.

app/models/user_mentee_application.rb
```diff
@@ -28,9 +28,29 @@
...
 
   validates :city, :state, :country, :reason_for_applying, :learned_to_code, :project_experience,
     :available_hours_per_week, presence: true
+
+  after_create :create_initial_application_state
+
+  delegate :rejected?, to: :current_state
+
+  def promote_application(user)
+    next_status = MenteeApplicationState.next_status(self)
```
afogel, line 43: this shouldn't accept an application, just a status.

app/models/mentee_application_state.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def self.next_status(mentee_application)
```
afogel, line 36: would suggest a refactor to look more like what I wrote, though am fine if it's an instance method rather than class method.

app/models/mentee_application_state.rb
```diff
@@ -0,0 +1,59 @@
...
+    current_state = current_state(mentee_application)
+
+    rails "Can't promote #{self}" unless can_promote?(mentee_application)
```
afogel, line 39: should be `raise`, not `rails`.

afogel (changes requested):

GALTdea:
@toyhammered I think this PR might be taking more time than we have. At this point, I'd appreciate it if you help us decide what methods need to be delegated to the mentee_application_state. 
I love how Ariel decoupled the next_status logic. But I think other than that, it'll just be a matter of  preference

toyhammered:
> @toyhammered I think this PR might be taking more time than we have. At this point, I'd appreciate it if you help us decide what methods need to be delegated to the mentee_application_state. I love how Ariel decoupled the next_status logic. But I think other than that, it'll just be a matter of preference

What i'd do is literally just copy/paste everything in this PR https://github.com/agency-of-learning/PairApp/pull/280/files into your branch. It is good enough for now and includes all the tests we need. Afterwards, make sure the flows still work and we can get it merged in. We can do refactoring later.

PR #264, opened 2023-08-27
add open_to_work banner to profile pic

afogel:
## What's the change?
replaces the existing banner with an svg overlay

only shows overlay in the event that profile is marked `open_to_work`

## Demo / Screenshots
Screen Shot 2023-08-27 at 1 26 24 PM

### DIFF
--- a/app/components/profile/picture_component.html.erb
+++ b/app/components/profile/picture_component.html.erb
@@ -1,6 +1,10 @@
 <figure class="rounded-full w-64 h-64 mx-auto sm:mx-0 relative border-2 border-black bg-black mb-2">
-  <%= image_tag profile_picture_src, class: 'w-full object-cover rounded-full' %>
-  <%= tag.figcaption(class: "absolute bottom-0 translate-y-1 z-90 py-2 rounded-lg w-full text-center font-bold #{figcaption_colors}") do %>
-    <%= profile.job_search_status.humanize.titleize %>
-  <% end %>
+  <div class="relative inline-block w-full object-cover rounded-full">
+    <%= image_tag profile_picture_src, class: 'h-auto max-w-full block' %>
+    <% if open_to_work? %>
+      <div class="absolute top-0 left-0 w-full object-cover rounded-full">
+        <%= render inline: Rails.root.join('app/assets/images/open_to_work.svg').read %>
+      </div>
+    <% end %>
+  </div>
 </figure>
--- a/app/components/profile/picture_component.html.erb
+++ b/app/components/profile/picture_component.html.erb
@@ -2,7 +2,7 @@
   <div class="relative inline-block w-full object-cover rounded-full">
     <%= image_tag profile_picture_src, class: 'h-auto max-w-full block' %>
     <% if open_to_work? %>
-      <div class="absolute top-0 left-0 w-full object-cover rounded-full">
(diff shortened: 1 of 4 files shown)


app/components/profile/picture_component.html.erb
```diff
@@ -1,6 +1,10 @@
 <figure class="rounded-full w-64 h-64 mx-auto sm:mx-0 relative border-2 border-black bg-black mb-2">
-  <%= image_tag profile_picture_src, class: 'w-full object-cover rounded-full' %>
-  <%= tag.figcaption(class: "absolute bottom-0 translate-y-1 z-90 py-2 rounded-lg w-full text-center font-bold #{figcaption_colors}") do %>
-    <%= profile.job_search_status.humanize.titleize %>
-  <% end %>
+  <div class="relative inline-block w-full object-cover rounded-full">
+    <%= image_tag profile_picture_src, class: 'h-auto max-w-full block' %>
+    <% if open_to_work? %>
+      <div class="absolute top-0 left-0 w-full">
+        <%= render inline: Rails.root.join('app/assets/images/open_to_work.svg').read %>
```
afogel, line 6: Dunno whether this is the most idiomatic way to render the svg inline, but seems like it works. Maybe could be pushed to the component, just wanted to get something up and running

JoshDevHub: I've most often seen people use [this gem](https://github.com/jamesmartin/inline_svg), but I'm not immediately opposed to this if it works well for this  one-off usecase. Not sure how many svgs we'll be using.

app/components/profile/picture_component.rb
```diff
@@ -26,4 +26,8 @@ def profile_picture_src
       'placeholder_profile_picture.png'
     end
   end
+
+  def open_to_work?
+    profile.job_search_status == 'open_to_work'
```
toyhammered, line 31: shouldn't you be able to do `profile.open_to_work?` because it is an enum. Argueable, you can now just use this directly in the `erb` file because it is part of the model anyways.

toyhammered (commented):
one small comment, but this looks great otherwise! I am fine with doing this one off svg w/o using a gem. If we end up needing more we can revisit it later.

app/components/profile/picture_component.rb
afogel: I thought so, but tried it earlier and it didn't work. I might've messed it up, tho, so will revisit tomorrow re-update if it doesn't complain

toyhammered (commented):
Sorry for the delay on this one. During our townhall we were not sure if we wanted to allow this image to be displayed on their user profile.

cc @dpaola2

afogel:
@toyhammered  & @dpaola2  -- should we remove the functionality altogether? Until it's removed, I think it makes more sense to merge now, since it's a significant improvement on how it currently works. In the event that we choose to completely remove from the codebase, we can remove it alongside the work then.

toyhammered:
> @toyhammered & @dpaola2 -- should we remove the functionality altogether? Until it's removed, I think it makes more sense to merge now, since it's a significant improvement on how it currently works. In the event that we choose to completely remove from the codebase, we can remove it alongside the work then.

I am fine merging in. The specific part we will need to remove is the actual logic for the banner. I still think the enum is valuable to have.

toyhammered (approved):

Issue #179, opened 2023-07-20
[Profiles] Resume Upload


afogel:
User should be able to upload and show their resume.
Resumes should be versioned, such that you can select a different version of your resume that you'd like to show.

Open questions for discussion: 
1. Should the resume be a link to click (open in a new window) or embedded on the page?
  i. If displaying inline, using an iframe? stimulus controller wrapping [pdf.js]( https://github.com/mozilla/pdf.js)?
2. What format should we require the resume upload (pdf or some form of image?)?

~~Blocked until #180 completed~~

PR #265, opened 2023-08-27
add resume upload

afogel:
## What's the change?
This PR enables users to upload and update their resumes. This PR also enables direct upload, such that files get stored directly on a storage service and then the attachments are created via signed_id rather than an io to the filename.

In this PR, given that the Resume lives on the user, but the resource is a Profile (and they are only related via User), I encapsulated the form using the form object pattern for validations/collecting attributes, and passed it to a service/command object to process the relevant business logic.

Finally, I reorganized the form so have a bit more logical groupings, making it easier to skim and parse. Though I don't think it's in scope here, I think we should probably redesign the profile page form.

## Demo / Screenshots
https://www.veed.io/embed/0667b780-b21b-4211-be27-8e58a82503e9

#179 

- [X] Have you thought of misfiring code? e.g. too many loops, n+1, or how to handle nils?
- [X] Any validations to data needed?
- [X] Can (and should) any models be extracted?
- [] Did you leave helpful inline PR comments for the reviewer(s)?
- [X] Do you have a plan for how this change should be rolled back? (e.g., how do you deal with a migration rollback? Is your feature released behind a feature flag?)


spec/commands/profiles_update_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+require 'rails_helper'
+
+RSpec.describe Profiles::Update do
+  subject(:command) { described_class.new(form:) }
+
+  let(:profile) { create(:profile) }
+  let(:user) { profile.user }
+  let(:form) { ProfileForm.new(profile:, user:) }
+  let(:uploaded_resume) do
+    ActiveStorage::Blob.create_and_upload!(
```
afogel, line 12: Use this b/c flow expects a direct upload rather than attaching the file via io.

app/views/profiles/edit.html.erb
```diff
@@ -1,11 +1,29 @@
 <h1 class="text-2xl text-center">Edit Profile</h1>
 
-<div class="max-w-prose px-2">
-  <%= form_with model: @profile do |form| %>
+<div class="max-w-prose px-2 mx-auto my-4">
+  <%= form_for @profile_form, format: :html do |form| %>
```
JoshDevHub, line 4: Does this type of pattern only work with `form_for`? Only asking because `form_for` is [considered soft-deprecated](https://guides.rubyonrails.org/form_helpers.html#using-form-tag-and-form-for)

afogel: ah, didn't realize! I'll move to `form_with`

app/commands/profiles/update.rb
```diff
@@ -0,0 +1,52 @@
+module Profiles
```
toyhammered, line 1: So tbh I am not a big fan of this type of abstraction for the most part. To solve one of my problems, the `call!` should either return nothing or raise an error. Alternatively, you could make it `call` which can return true/false, but I much prefer raising custom errors specific to what went wrong.

(Another solution would be to pass in the controller in via self).

Now, given that we will be having a resume upload needed in the `user_mentee_application` I think tailoring to that specifically, using a service object is acceptable. That being said, what should be passed in here is most likely a `user` and a `resume`.

There might be some complications with figuring out how to properly redirect, but we can figure that out later.

app/forms/profile_form.rb
```diff
@@ -0,0 +1,35 @@
+class ProfileForm
```
toyhammered, line 1: Given what I said above about extracting the resume to it's own service object, I don't think we really need this. Most the complications came because of that, and everything else is just simple rails form stuff which should be in the model.

app/views/profiles/edit.html.erb
toyhammered: learn something new every day, didn't either know this lol

toyhammered (changes requested):
I've left some comments and thoughts. Let me know what you think and if what I said makes sense.

app/forms/profile_form.rb
afogel: How do you typically approach handling form data when it's not associated with a particular resource on the unrelated resource's controller? Do you namespace it differently in the form? Do you create a separate `params` method that requires those data, e.g. 
```ruby
def resume_params
  params.require(:profile).permit(:resume, :resume_name, :current_resume_id)
end
```

app/commands/profiles/update.rb
afogel: When you have some time, I'd be curious to hear a bit more about the time using this type of command/service object abstraction to encapsulate domain logic ended up causing an issue for you.

I'm also unclear about why you prefer `call!` to return nothing or raising an error? I can understand wanting to bubble up an error, though as long as that error gets handled and reported (we probably could look into rollbar, honeybadger, or something similar), seems like gracefully recovering from it is important.

I have historically referenced the [stitchfix post on anatomy of a service object](https://multithreaded.stitchfix.com/blog/2015/06/02/anatomy-of-service-objects-in-rails/) as rough guidelines and specifically like returning rich objects. In this case, I chose to use symbols over a more complex object to simplify the implementation so I could solicit feedback more quickly. Additionally, it does mirror other libraries that follow similar patterns, like [rectify](https://github.com/andypike/rectify).  However, one thing I like about the symbols is that they they provide more context about what the method did rather than silently passing, raising an error, or true/false.

I actually do like failing noisily, so I don't mind rescuing and reraising if the Error is not what I expect it to be, but am wary of it until it's clear that we have a solid error reporting system in place that doesn't completely disrupt service to the end users.

app/forms/profile_form.rb
toyhammered: I don't have too much of an opinion. separate params would be totally fine by me though

app/commands/profiles/update.rb
toyhammered: Whenever i've done service objects I generally follow the `call!` pattern (cause I like to fail noisily). (There are times when i'd have different methods if different places need different data from the same source).

My thought process goes like this, the `!` means it will raise some error on failure and it is your job as the caller to catch it and handle appropriately. If it returns nothing, that means the service succeeded in what it was doing (this can also mean the service didn't need to do anything because making them idempotent is nice). 

As for reporting, yeah we should prob implement that at some point. I generally tend to use Sentry and have had good results. 

Sidenote:

Now we could include Dry Structs/ Dry types or ways to returned typed data to keep things consistent but I still don't see much value adding more levels of complexity.

app/forms/profile_form.rb
afogel: @toyhammered -- since there's no attribute for resumes on the `Profile` model, it becomes problematic to leverage rails form helpers. The `ProfileForm` addresses this issue.
I can still update so that the command is only updating the resume, but I'm going to push back on removing the Form Object here.

afogel:
Responding to @toyhammered 's comments, I've now updated the PR to remove the form object. One outstanding concern I have is handling validation errors on the resume side of things, since that's not technically covered by the profile errors. 
That said, service object encapsulates the more complex logic associated with updating a resume and is fully tested, so maybe that's less problematic.

I've also moved the business logic associating with updating the current profile and current resume into the profile controller and wrapped all of it in a single transaction block.

app/commands/resumes/update.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def initialize(user:, params:)
+      @user = user
+      @params = cast_to_attributes(params)
+      @current_resume_id = params[:current_resume_id]
+    end
+
+    def call!
+      ActiveRecord::Base.transaction do
+        # rubocop:disable Rails::SkipsModelValidations
```
afogel, line 13: disabling this linter check that forbids `update_all` here because I've wrapped the database logic in a single transaction. Since subsequent database calls should fail noisily, there should always be a current resume that is assigned or the entire transaction should be rolled back.

app/models/user.rb
```diff
@@ -72,8 +72,11 @@ class User < ApplicationRecord
   has_many :blog_posts, dependent: :destroy
 
   has_one :profile, dependent: :destroy
-  has_many :resumes, dependent: :destroy
+  has_many :resumes, dependent: :destroy, inverse_of: :user
   has_one :mentee_application, class_name: 'UserMenteeApplication', dependent: :destroy
+  # rubocop:disable Rails/InverseOf
+  has_one :current_resume, -> { where(current: true) }, class_name: 'Resume', dependent: nil
```
toyhammered, line 78: shouldn't this by `find_by(current: true)` because it is a has_one?

app/commands/resumes/update.rb
```diff
@@ -0,0 +1,34 @@
...
+      ActiveRecord::Base.transaction do
+        user.current_resume.update!(current: false)
+
+        if attributes[:resume]
+          user.resumes.create!(attributes)
+          next
+        end
+
+        user.resumes.find_by(id: current_resume_id).update!(current: true)
+      end
```
toyhammered, line 21: I think we can optimize this quickly a tiny bit. We should be able to `next` if `attributes[:resume].blank?` && current_resume.id == current_resume id.

Other than that this makes sense what is happening. Might be worth adding a small comment explaining that they can chose a different resume to make as current (took me a min to figure that out), and not need to upload one.

app/controllers/profiles_controller.rb
```diff
@@ -1,16 +1,19 @@
...
+      @profile.update!(profile_params)
+      Resumes::Update.new(user: current_user, params: resume_params).call!
+    end
+    if update_records
       redirect_to @profile, success: 'Profile successfully updated!'
     else
+      # using this now misses validations and errors on the resume part of the form.
       flash.now[:form_errors] = @profile.errors.full_messages
       render :edit, status: :unprocessable_entity
     end
```
toyhammered, line 19: I feel like it should always just redirect and the `else` can just happen if there is an error (i.e: rescue).

toyhammered (changes requested):
Looking good, just couple small things!

app/models/user.rb
afogel: In general, you're correct, but when I used `find_by`, it throws up an exception. I think it's the way that the scope using a `has_one` functions. Even the documentation on `has_one` shows it that way:
https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_one-label-Scopes

app/commands/resumes/update.rb
afogel: I think some comments might make sense to clarify some, but I think the logic flow may be confusing, not least of which because `next` is not really a correct word for the context, but rubocop is irritating (Rails/TransactionExitStatement). I can update to ignore rubocop in this context.

It should be this:
```ruby
  ActiveRecord::Base.transaction do
    user.current_resume.update!(current: false)
    return user.resumes.create!(attributes) if attributes[:resume]
    user.resumes.find_by(id: current_resume_id).update!(current: true)
  end
```

app/controllers/profiles_controller.rb
afogel: Can you clarify this, please?

app/commands/resumes/update.rb
toyhammered: what rubocop is saying is correct. I understand the `next` and `return` difference. It was more asking if we could include a guard clause because `user.current_resume.update!(current: false)` and `user.resumes.find_by(id: current_resume_id).update!(current: true)` might be the exact same record which could already be true.

app/controllers/profiles_controller.rb
toyhammered: this service isn't going to be returning anything. Wouldn't it just be `nil` or raising some type of error which you'd need to capture in the controller?

afogel: got it. yeah, fair enough.

app/commands/resumes/update.rb
afogel: gotcha. updated.

toyhammered (approved):

PR #266, opened 2023-08-27
Add profile links

### LINKED ISSUE #187: [Profiles] Add Profile Links
Add profile links for different sites (LinkedIn, Discord, Github, Personal Webiste)

JoshDevHub:
## What's the change?
- Add fields for profile links to `Profile`
- Allow users to edit their links
- Add component for displaying profile links on the users' profile pages

## Highlights / Risks / Surprises / Cleanup
Migrate down to cleanup

Viewing and editing user profiles

#### Desktop
image

#### Mobile
image

#### Hover effect:
profile_links_hover

I just thought this looked nice. Glad to remove it if others don't though!

Closes #187 

### DIFF
--- a/app/components/profile/link_component.rb
+++ b/app/components/profile/link_component.rb
@@ -2,10 +2,10 @@
 
 class Profile::LinkComponent < ViewComponent::Base
   LINK_ICONS = {
-    github_link: 'fa-brands fa-github text-[#181717] hover:bg-[#181717]',
-    linked_in_link: 'fa-brands fa-linkedin-in text-[#0072b1] hover:bg-[#0072b1]',
+    github_link: 'fa-brands fa-github text-neutral hover:bg-neutral',
+    linked_in_link: 'fa-brands fa-linkedin-in text-linkedin-brand hover:bg-linkedin-brand',
     personal_site_link: 'fa-solid fa-link hover:bg-neutral',
-    twitter_link: 'fa-brands fa-twitter text-[#26a7de] hover:bg-[#26a7de]'
+    twitter_link: 'fa-brands fa-twitter text-twitter-brand hover:bg-twitter-brand'
   }.freeze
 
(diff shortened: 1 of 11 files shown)


app/components/profile/link_component.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+class Profile::LinkComponent < ViewComponent::Base
+  LINK_ICONS = {
+    github_link: 'fa-brands fa-github text-[#181717] hover:bg-[#181717]',
+    linked_in_link: 'fa-brands fa-linkedin-in text-[#0072b1] hover:bg-[#0072b1]',
+    personal_site_link: 'fa-solid fa-link hover:bg-neutral',
+    twitter_link: 'fa-brands fa-twitter text-[#26a7de] hover:bg-[#26a7de]'
```
JoshDevHub, line 8: So I know Ariel had mentioned wanting to avoid the `[hex-code]` color syntax, which I absolutely agree is a good idea. Not sure about these 'brand colors' though as they're not part of our app's theme, so I've treated them as an exception here.

Do  you have any opinions on this @afogel ?

afogel: this looks great! I do have some opinions haha

I think it might make sense to add css variables for the brand colors/add a theme color in the tailwind.config. My thinking is along these lines:
Not only will it be more readable/easier to understand what the colors are, but it will continue to enforce a coding style wherein we don't have one-off colors that are added. My biggest concern is leaving code examples for future developers that may be less initiated with our norms.

I think we probably could get away with using our neutrals for the github, though?

afogel (approved):
on balance, looks great. My comment isn't blocking, but I think it could be a nice way to reinforce norms. Having a thoughtful conversation about when we add new variables to the theme class seems like a worthwhile use of team energy/time.

PR #282, opened 2023-09-02
rake task to backfill applicants using csv data

afogel:
This is a rake task that backfills the applicants.

A few things to note:
datasets needed to be cleaned. In particular, cleaning and parsing location took some time. On the existing form, I would update the prompt to ask people just put in a location, nothing else (e.g. "London, UK" instead of "From Boston, but I moved to London earlier this year").

Dataclips dataset doesn't have hours/week, even though it's a required category on the mentee application model. Not sure where the dataclips data are being sourced, but I'd update the form if it's still relevant to ask that information.

Responses csv doesn't have github links.


lib/tasks/backfill_applicants.rake
```diff
@@ -0,0 +1,88 @@
...
+    email = row['email']
+
+    city, state, country = grab_location(row["location"])
+    additional_information = row["additional_information"]&.squish
+    reason_for_applying = row["reason_for_applying"]&.squish
+    linkedin_url = row["linkedin_url"]
+    learned_to_code = row["learned_to_code"]&.squish
+    project_experience = row["project_experience"]&.squish
+    # hours/week not asked on dataclips file, so default to 15
+    available_hours_per_week = 15
```
toyhammered, line 66: what we should do for this is after it creates the `UserMenteeApplication` (we should store that in a variable and then update it to be 0 with a `validate: false`. This way when they reapply (because it will just shove all the details into the form and they will have resubmit, it will force them to fill that out.

lib/tasks/backfill_applicants.rake
```diff
@@ -0,0 +1,88 @@
...
+  uri = URI("https://nominatim.openstreetmap.org/search?q=\"#{location_string}\"&format=json&&addressdetails=1&accept-language=en")
+  response = Net::HTTP.get_response(uri)
+  location_data = JSON.parse(response.body)[0]["address"]
+  
+  [location_data["city"], location_data["state"], location_data["country"]]
+end
+
+task backfill_applicant_responses: :environment do
+  headers = ["#","name","email","reason_for_applying","linkedin_url","learned_to_code","location","project_experience","available_hours_per_week","referral_source","additional_information","start_utc","submit_utc","network_id","tags"]
+  table = CSV.read(Rails.root.join('lib', 'tasks', 'responses.csv'), headers:,  liberal_parsing: true)
```
toyhammered, line 16: Assuming your plan is to:
1. ssh in
2. rails c
3. create a file with this name and copy the data into it (via console)
4. consume

toyhammered: Alternatively what you can do is run this locally and export the data (in the table) as json (well link it via email or something). This way we only need to create account and then jsut copy/paste json into variable and practically consume. Your call though in how you'd want to do this just throwing out another idea.

toyhammered (commented):
Love it! Left some ideas that I wanted to hear your thoughts on but overall this looks great.

lib/tasks/backfill_applicants.rake
afogel: sure -- that works, too.

lib/tasks/backfill_applicants.rake
afogel: yeah, I figured that I'd tackle that implementation detail once I got others' eyes on the core task that works, especially b/c there are some data integrity issues that I figure we should all be aware of. Fundamentally, though, it's a small modification, assuming the rest of the script works (which it does, locally, w/ the cleaned datasets).

toyhammered:
Also forgot to mention this. Josh is working on Cohorts https://www.notion.so/toyhammered/UserMenteeApplication-Group-Cohort-8fe7e4ca3f4b468c8c284ff75662c7bd?pvs=4 and for these users we would want to add them to one called "original" or something (it will be inactive). This will allow them to then easily reapply when we reopen.

afogel:
k -- I'll wait for that work to be merged in, then revise this work. I'll probably create a cohort for Feb through October, and set it to inactive.

@toyhammered @dpaola2 -- are both csvs `responses`, and `dataclips` part of the original cohort?

> Also forgot to mention this. Josh is working on Cohorts https://www.notion.so/toyhammered/UserMenteeApplication-Group-Cohort-8fe7e4ca3f4b468c8c284ff75662c7bd?pvs=4 and for these users we would want to add them to one called "original" or something (it will be inactive). This will allow them to then easily reapply when we reopen.

toyhammered:
> k -- I'll wait for that work to be merged in, then revise this work. I'll probably create a cohort for Feb through October, and set it to inactive.
> 
> @toyhammered @dpaola2 -- are both csvs `responses`, and `dataclips` part of the original cohort?
> 
> > Also forgot to mention this. Josh is working on Cohorts https://www.notion.so/toyhammered/UserMenteeApplication-Group-Cohort-8fe7e4ca3f4b468c8c284ff75662c7bd?pvs=4 and for these users we would want to add them to one called "original" or something (it will be inactive). This will allow them to then easily reapply when we reopen.

yeah all part of original, just shove them together.

lib/tasks/backfill_applicants.rake
```diff
@@ -0,0 +1,88 @@
...
+    city, state, country = grab_location(row["location"])
+    additional_information = row["additional_information"]&.squish
+    reason_for_applying = row["reason_for_applying"]&.squish
+    linkedin_url = row["linkedin_url"]
+    learned_to_code = row["learned_to_code"]&.squish
+    project_experience = row["project_experience"]&.squish
+    # extract and coerce hours from string
+    available_hours_per_week = row["available_hours_per_week"]&.split(" ")&.find { |el| el.to_i > 0 }&.to_i
+
+    ActiveRecord::Base.transaction do
```
toyhammered, line 31: Also one more thing I thought of. We should probably set all the MenteeApplicationState to rejected (because that is going to most likely be needed for the reapply flow)

afogel:
For the purposes of future documentation and auditing, I'm leaving this comment that was sent directly to @toyhammered here on the PR as well.

The process of pasting in a JSON via the CLI wasn't really going to work -- when I tried that, copying from the buffer seemed slow and did not instill confidence in me that it was completed or working as intended.

I also considered SCP, but because of firewalls and trying to reverse engineer heroku's VPS system, it seemed like much more work than it was worth.

I've decided to upload the cleaned datasets to a secret gist that will be destroyed after the backfill is complete. I have updated the rake task to consume the response body from the raw gist and pushed this updated version (along with other requested changes) to the branch on origin, though I have redacted the URLs containing the gists. 

The updated script has been tested locally, first using the rake task, then directly inside the rails console.

afogel:
Tasks is complete

PR #283, opened 2023-09-04
Create User Mentee Application Cohorts

### LINKED ISSUE #271: [UserMenteeApplication] Add UserMenteeApplicationGroup
Add a new table called `UserMenteeApplicationGroup` (maybe `UserMenteeApplicationCohort`). The purpose of this feature is to be able to keep track of when a user submitted their application and in addition allow for easy reapply. (The reapplies feature will happen in a different feature)

The UserMenteeApplicationGroup should be assigned when a user submits an application. There should only ever be 1 active UserMenteeApplicationGroup at a time (tbd if that should be a boolean or date range). 

Don't worry about the client side for this. For now we will handle things via the console.

JoshDevHub:
## What's the change?
- Add new model for `UserMenteeApplicationCohorts`
- Use `after_save` callback to deactivate other cohorts if current one is active
- Associate new applications with currently active cohort

No practical effect in the app. For now this is intended to be interfaced with in the console only.

Migrate down to cleanup

Closes #271 

### DIFF
--- a/app/models/user_mentee_application_cohort.rb
+++ b/app/models/user_mentee_application_cohort.rb
@@ -15,6 +15,10 @@ class UserMenteeApplicationCohort < ApplicationRecord
 
   after_save :deactivate_other_cohorts!, if: :active?
 
+  def self.active
+    find_by(active: true)
+  end
+
   private
 
   def deactivate_other_cohorts!
--- a/app/models/user_mentee_application_cohort.rb
+++ b/app/models/user_mentee_application_cohort.rb
@@ -21,9 +21,9 @@ class UserMenteeApplicationCohort < ApplicationRecord
 
   private
 
+  # rubocop: disable Rails/SkipsModelValidations
   def deactivate_other_cohorts!
-    UserMenteeApplicationCohort.excluding(self).find_each do |cohort|
-      cohort.update!(active: false)
-    end
+    UserMenteeApplicationCohort.excluding(self).update_all(active: false)
   end
+  # rubocop: enable Rails/SkipsModelValidations
 end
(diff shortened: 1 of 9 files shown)


app/models/user_mentee_application_cohort.rb
```diff
@@ -0,0 +1,29 @@
...
+  after_save :deactivate_other_cohorts!, if: :active?
+
+  def self.active
+    find_by(active: true)
+  end
+
+  private
+
+  def deactivate_other_cohorts!
+    UserMenteeApplicationCohort.excluding(self).find_each do |cohort|
```
toyhammered, line 25: can we just do something like `UserMenteeApplicationCohort.excluding(self).update_all(active: false)`?

toyhammered (approved):
One small comment but this looks great! I kept on contemplating related to the `optional: true` part but I think I am fine with it for now.

app/models/user_mentee_application_cohort.rb
JoshDevHub: We could. Rubocop warns for use of `update_all` because validations don't run. I could of course just tell it to ignore this method. Don't think that would ever cause a problem in this case.

afogel: I wonder - if we did it in a single transaction, does that give the benefit we're looking for while also maintaining validations?

toyhammered: certain cases I am fine with using update_all because it is just faster. That generally goes when enabling/disabling something or mass updating. This is one of the cases where it is fine.

toyhammered (changes requested):
I realized one thing was missing. The cohort should have a name field (and maybe a display_name for client facing)?

toyhammered (approved):

PR #291, opened 2023-09-08
scope pair requests to active members

### LINKED ISSUE #290: Pair Request invitee dropdown menu showing none active members
**Describe the bug**
After we migrated user_mentee_application data from the old database, the invitee drop-down menu displays all users in the database with no regard for their membership status. 

We need to create a scope only to display members with an active_member status

**To Reproduce**
1. Go to 'https://pair.agencyoflearning.com/pair_requests'
2. Click on 'invitee dropdown menu'

**Expected behavior**
This dropdown should only render active members

afogel:
## What's the change?
Bug fix. Scopes pair request users to active members. Adds a `members` scope.
 
Makes pair requests more usable again.

closes #290 . Test by checking out the pair request index page and selecting the dropdown. You should only see members or admin.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -83,6 +83,8 @@ class User < ApplicationRecord
 
   after_create :create_profile!
 
+  scope :members, -> { where(role: :member) }
+
   def self.invite!(attributes = {}, invited_by = nil, options = {}, &)
     default_name = { first_name: 'First', last_name: 'Last' }
     super(attributes.merge(default_name), invited_by, options, &)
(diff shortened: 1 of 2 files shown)


JoshDevHub (approved):
Looks good.

Non-blocking, idle thought: I wonder if it's worth rethinking the "member" role name for users. If we're going to say that regular users and admins are both "members", but then also call the regular user role "member" -- this is weird semantics to me.

Like right now, you can say `User.member` and look up users with the member role (no admins) and say `User.members` and look up users with the member or admin role.

But anyways, this is a discussion that needs to happen separately from this PR probably. Approved :rocket:

PR #298, opened 2023-09-10
update navbar based on user role.

toyhammered:
## What's the change?
The navbar was not correct depending on what role you were assigned. I have done an update to make sure they are consistent across the roles now.

We already have another ticket to actually refactor the logic, this is just a temporary solution.


app/views/landing/index.html.erb
```diff
@@ -1,6 +1,5 @@
 <div class="bg-neutral lg:w-3/4 mx-auto rounded-lg text-neutral-content">
   <div class="sm:px-[5em]">
-    <%= render_navbar %>
```
afogel, line 3: This completely removes the the navbar from the landing page, right? Screenshots would be helpful here.

afogel: ah, got it; just leverages the `layouts/_navbar` partial, no matter where you are in the app.

afogel (approved):
Screenshots would still be helpful, but lgtm. Small incremental change that seems like it'll both support and force the issue to refactor the navbar (and possibly the landing page color scheme).

PR #301, opened 2023-09-12
Application index with reapply

### LINKED ISSUE #272: [UserMenteeApplication] Allow users to reapply if they were previously rejected
If a user was previously rejected, allow them to reapply (if there is a new UserMenteeApplicationCohorts available that they are not part of). The way this would work is that it would take all the existing information and shove it into the form. They will be able to review/edit the form and then click submit (which will reapply them).

This is just a shortcut to allow them to easily reuse their information. They could in theory, just apply again and put in any information.

JoshDevHub:
## What's the change?
- Allow users to have many applications. Refactor parts of the application flow to accommodate this.
- Add application index page for applicant users
- Allow applicants to reapply if there's a new active cohort that they haven't applied to yet.

Signing up and logging in for applicant users now directs to the applications index page. We could think about doing this differently. Part of the reason I did it was to provide something for people who sign up when there's no active cohort (because filling in an application wouldn't make sense).

Applicant users viewing their applications and applying/reapplying to them.

#### Mentee Applications Index with No Applications (Apply CTA)
Selection_289

#### Mentee Applications Index with Application In-Review
Selection_290

#### Mentee Applications Index with Reapply CTA
Selection_291

Closes #272 


app/components/user_mentee_application/list_item_component.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  private
+
+  attr_reader :mentee_application
+
+  def link_text
+    application_cohort = mentee_application.user_mentee_application_cohort
+
+    "Application for #{application_cohort&.name || 'The Agency of Learning'}"
```
JoshDevHub, line 15: Have to occasionally do these contortions to account for a possible null cohort. Would maybe be cool to have cohorts be mandatory for any application, but I can't remember if that effects some of the other ideas around this feature.

app/components/user_mentee_application/reapply_cta_component.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def cta_action_text
+    last_application.present? ? 'Reapply' : 'Apply'
+  end
+
+  def application_not_in_cohort?
+    return true if last_application.blank?
+
+    application_cohort_id = last_application.user_mentee_application_cohort_id
+    application_cohort_id.blank? || (application_cohort_id != active_cohort.id)
```
JoshDevHub, line 25: Seems like I can use `UserMenteeApplicationCohort#application_in_cohort?` here, but since it calls `#exists?` it *will* hit the db even if all the data to determine this is loaded.

Generally like to avoid db hits from components, although it might be worth noting that it wouldn't lead to N+1 in this situation.

app/components/user_mentee_application/list_item_component.rb
toyhammered: cohorts are mandatory for any application (or should be)

toyhammered: Ok just checked and we have backfilled everything. Feel free to make it `null: false` in this PR.

app/components/user_mentee_application/reapply_cta_component.rb
toyhammered: good shoutout and what you have here is fine.

toyhammered (changes requested):
This is fantastic! Just left you one comment related to making the cohorts required.

app/components/user_mentee_application/list_item_component.rb
```diff
@@ -12,6 +12,6 @@ def initialize(mentee_application:)
   def link_text
     application_cohort = mentee_application.user_mentee_application_cohort
 
-    "Application for #{application_cohort&.name || 'The Agency of Learning'}"
+    "Application for #{application_cohort.name}"
```
afogel, line 15: 🔥

toyhammered (approved):
Great job!

PR #303, opened 2023-09-12
add ERD to interact with data model

afogel:
## What's the change?
Integrates a gem that will create a PDF of the application's ERD. Will update the ERD when migrations are run.
This can be used as both a tool for thought, as well as a way to help new developers onboard to the PairApp.

Creates a new version of the ERD every time a migration is run.

Requires Graphviz 2.22+.
On mac: `brew install graphviz`
On Linux: `sudo apt-get install graphviz`

## Demo / Screenshots
see [erd.pdf](https://github.com/agency-of-learning/PairApp/blob/7c9c58d83962ac4a55a81811b3efd1921c23cf84/erd.pdf)


toyhammered (approved):
just fix linting but love it! We can prob do as an iterative PR to write to S3 via actions and have that become a public url.

Issue #226, opened 2023-08-11
[UserMenteeApplication] We should be notified when a new application is submitted


toyhammered:
When someone submits an application, we should be notified by email that an application has been submitted.
We should also send an email to the user telling them that we have received their application and linking them to the show page (user mentee application). We should also send a notification to the admin to let them know a new application has been submitted.

GALTdea:
@toyhammered I can tackle this issue

PR #302, opened 2023-09-12
294 build state machine for application flow

afogel:
## What's the change?
This introduces the state machine to handle transitions through the mentee application flow, based on the flow laid out in the [notion document](https://www.notion.so/toyhammered/Application-Submission-Design-Doc-7a45fd8e25ce4f448e9b47bc00a142b5).

The existing ServiceObject conforms to the API laid out in the [notion design doc](https://www.notion.so/toyhammered/Applicant-State-Machine-Design-Doc-eb61b1dc2ae4417da53ada037cea3916) that I published yesterday.

## What key workflows are impacted?
Currently, there are none. In the next PR I upload, I will hook it up to the existing forms/controllers

## Highlights / Surprises / Risks / Cleanup

1) we added an addition state the enum for MenteeApplicationState, so I've included a migration so that the change doesn't impact the mapping for existing applications that have already gone through the funnel
2) @dpaola2 @toyhammered I specifically want to call attention to the copy of the new mailers that will get sent off for acceptances/rejections. 

For rejections, I used the more verbose copy I drafted since it reflects the YC-style reapplication ethos while using a tone of voice more reflective of AOL's culture.

For Acceptances, I wasn't clear on the appropriate CTAs that we want to communicate after we notify applicants that they were accepted. Please let me know what I can tweak

3) I assume that we probably want the state machine to handle sending out the notification email with the code challenge, but for now I've omitted it since it was not clear. If that falls within scope, I will update to also send a notification when promoted to `:code_challenge_sent`.

## Issue ticket number and link
#294 

## Checklist before requesting a review

Please delete items that are not relevant.

- [X] Did you add appropriate automated tests?
- [X] Did you consider risks around security, performance, etc.?
- [X] Have you thought of misfiring code? e.g. too many loops, n+1, or how to handle nils?
- [X] Any validations to data needed?
- [X] Related to readability and consistency: Did you add comments and/or documentation? (README, Notion, etc.)
- [X] Is everything elevated to the highest level? (e.g., can logic be moved out of view into controllers, or out of controllers into models?)
- [X] Can (and should) any models be extracted?
- [X] Is there any linting that needs to be executed?
- [X] Did you leave helpful inline PR comments for the reviewer(s)?
- [X] Do you have a plan for how this change should be rolled back? (e.g., how do you deal with a migration rollback? Is your feature released behind a feature flag?)

PR #304, opened 2023-09-13
Mentee application notifications

GALTdea:
## What's the change?
Adds email notification to email applicants when submitting mentee_applications 
Adds email notifications to alert admins when a new mentee_application is submitted 

#226 


app/mailers/user_mentee_application_alert_mailer.rb
```diff
@@ -0,0 +1,7 @@
+class UserMenteeApplicationAlertMailer < ApplicationMailer
+  def notify_for_application_submission
+    @user_mentee_application = params[:user_mentee_application]
+    @user = User.admin.first
+    mail(to: @user.email, subject: 'Mentee Application Submitted')
```
JoshDevHub, line 5: So a couple of things here:

If you look in `ApplicationMailer`, you'll see that mailers inherit a `set_recipient` method and a `before_action` filter that calls it. Noticed classes automatically pass the mailer the `recipient` param for whoever the notification is being delivered to.

`ApplicationMailer` also specifies that the default `to` is for the recipient's email address. This all basically just allows us to write these emails with less repetition. This method could just become
```rb
def notify_for_application_submission
  @user_mentee_application = params[:user_mentee_application]
  mail(subject: 'Mentee Application Submitted')
end
```
And then in the template, you could refer to `@user_mentee_application` if you need data from that (which you're already doing), but you'll also have `@recipient` to refer to the user that is receiving the email.

And same stuff goes for the other mailer. Definitely let me know if any of this is confusing.

spec/mailers/previews/user_mentee_application_alert_preview.rb
```diff
@@ -0,0 +1,3 @@
+# Preview all emails at http://localhost:3000/rails/mailers/user_mentee_application_alert
+class UserMenteeApplicationAlertPreview < ActionMailer::Preview
```
JoshDevHub, line 2: You can get rid of these previews. I think there was a brief discussion about introducing them later, but for now, we're not using them.

spec/mailers/user_mentee_application_alert_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe UserMenteeApplicationAlertMailer do
+  pending "add some examples to (or delete) #{__FILE__}"
```
JoshDevHub, line 4: I would say I'm in favor of testing the mailers. They're inexpensive to write and maintain. And they're useful for catching potential problems. I know the existing mailers don't have coverage yet. Maybe that's a good tech debt ticket.

Ariel has written some mailer specs for his [PR here](https://github.com/agency-of-learning/PairApp/pull/302). You can look at it to get some ideas for how to test them.

JoshDevHub (changes requested):
Left a few comments/thoughts. Let me know if you have any questions!

app/controllers/user_mentee_applications_controller.rb
```diff
@@ -61,4 +63,12 @@ def user_mentee_application_params
   def resume_params
     params.require(:user_mentee_application).permit(:resume, :resume_name)
   end
+
+  def send_notifications
+    UserMenteeApplication::ApplicationSubmissionNotification
+      .with(user_mentee_application: @user_mentee_application).deliver(@user_mentee_application.user)
+
+    UserMenteeApplication::ApplicationSubmissionAlert
+      .with(user_mentee_application: @user_mentee_application).deliver(User.admin)
```
toyhammered, line 72: For the deliver, you can send to both Dave and I, that should be enough.

toyhammered (commented):
overall this looks great! Think there are just a few small pieces that need to be done.

spec/mailers/previews/user_mentee_application_alert_preview.rb
afogel: I found them really helpful so that I could do quick iteration on the styling of the message without having to send a message and see how it looks.
That said, I agree that if you haven't actually filled it out and loaded in the template, you should get rid of them. Empty previews aren't helpful.

app/notifications/user_mentee_application/application_submission_alert.rb
```diff
@@ -0,0 +1,7 @@
+class UserMenteeApplication::ApplicationSubmissionAlert < Noticed::Base
+  deliver_by :email,
+    mailer: 'UserMenteeApplicationAlertMailer',
+    method: 'notify_for_application_submission'
```
afogel, line 4: we've been a bit lax about this on other notifications in the code base, but I think it's important that work like mailers that can be performed asynchronously is, in fact, performed asynchronously. The only way I could confirm that it is processed via ActiveJob and properly enqueued is to add the `enqueue: true` option to the `deliver_by` method.

app/controllers/user_mentee_applications_controller.rb
afogel: lol:
```ruby
# user.rb

scope :dave, ~> { find_by(email: 'dave@aol.com') }
```
Probably better might be like, but we probably want some form of encapsulation for that, right?

```ruby
# user.rb

scope :team_leads, ~> { where(email: ['daniel@agencyoflearning.com','dave@agencyoflearning.com']) }
```

app/controllers/user_mentee_applications_controller.rb
```diff
@@ -61,4 +63,12 @@ def user_mentee_application_params
   def resume_params
     params.require(:user_mentee_application).permit(:resume, :resume_name)
   end
+
+  def send_notifications
+    UserMenteeApplication::ApplicationSubmissionNotification
```
afogel, line 68: In the [best practices](https://github.com/excid3/noticed#-best-practices) for the noticed gem, it suggests defining these notifications on the model and leveraging callbacks. I think this would be a good candidate for moving these methods out of the controller and instead defining them on application model, invoking them with `after_create_commit`s.

This would also prevent issues with notifications being sent, even if the transaction doesn't succeed.

afogel (changes requested):
big 2 requests for change from me are to make sure that mailer work is performed asynchronously and to define the notifications on the model and invoke using after create callbacks. Great start!

app/controllers/user_mentee_applications_controller.rb
JoshDevHub: I interpret that section of the README not as a recommendation to use callbacks for these notifications, but as a recommendation to specifically use `after_create_commit` when you *are* using callbacks.

That said I think I might still be in favor of doing this via callback because having it in the controller means the notification side effects go untested.

app/mailers/user_mentee_application_alert_mailer.rb
GALTdea: Oh that's awesome! changes made!

spec/mailers/user_mentee_application_alert_spec.rb
GALTdea: agree! done

app/controllers/user_mentee_applications_controller.rb
afogel: Lol, josh gets the careful reading award 😊  I actually agree that I don't think they should be in callbacks every time, but I'll advocate for it here still both for the testing reason as well as because the controllers may send them out despite an application failing to get created.

app/notifications/user_mentee_application/application_submission_alert.rb
GALTdea: hmm, I think this is a good idea, but for the sake of time, I'll probably submit a different PR for it

app/controllers/user_mentee_applications_controller.rb
GALTdea: done! good point @afogel.

app/controllers/user_mentee_applications_controller.rb
GALTdea: Are these emails accurate? 😆  but yeas, i agree with this

JoshDevHub (changes requested):
A few more changes.

Another thing: the coverage interprets the after create callback as being "tested" because there are already tests [written for `#create`](https://github.com/agency-of-learning/PairApp/blob/main/spec/models/user_mentee_application_spec.rb#L38-L46)

But these existing tests aren't verifying the behavior of the new callback. I'd be in favor of adding some tests to `#create` that check that the notification emails are being sent on create.

You can get an idea for how this is done by looking at the service object tests in Ariel's currently open PR. [Here's a link to what I mean.](https://github.com/agency-of-learning/PairApp/pull/305/files#diff-64e1794cd8eb5323e9d25c3017b5c0710be68824d447de5c307e4a7b6ef0f8faR83-R87)

app/mailers/user_mentee_application_mailer.rb
```diff
@@ -0,0 +1,6 @@
+class UserMenteeApplicationMailer < ApplicationMailer
+  def notify_for_application_submission
+    @user_mentee_application = params[:user_mentee_application]
```
JoshDevHub, line 3: Do you even need the `@user_mentee_application` in this mailer? Doesn't seem the template is using it.

afogel (approved):
Once Josh's suggestions are addressed, I think this is good to go!

app/mailers/user_mentee_application_alert_mailer.rb
```diff
@@ -0,0 +1,6 @@
+class UserMenteeApplicationAlertMailer < ApplicationMailer
+  def notify_for_application_submission
+    @user_mentee_application = params[:application]
```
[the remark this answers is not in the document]
GALTdea: oops, typo.. fixed

app/mailers/user_mentee_application_mailer.rb
GALTdea: actually, I wanted to add a link to the application in the mailer template like this ```  
<p>While you wait, feel free to <%= link_to 'explore your application', user_mentee_application_url(@user_mentee_application) %> for more educational opportunities.</p> ``` 
but for whatever reason rails is not recognizing the route 🤷‍♂️ ...

app/models/user_mentee_application.rb
```diff
@@ -46,6 +46,7 @@ class UserMenteeApplication < ApplicationRecord
     :available_hours_per_week, presence: true
 
   after_create :create_initial_application_state
+  after_commit :send_notifications
```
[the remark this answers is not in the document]
GALTdea: done!

app/models/user.rb
```diff
@@ -86,6 +86,7 @@ class User < ApplicationRecord
   after_create :create_profile!
 
   scope :members, -> { where(role: %i[member admin]) }
+  scope :super_admins, -> { where(email: ['daniel@agencyoflearning.com', 'dave@agencyoflearning.com']) }
```
GALTdea, line 89: I added this scope to be able to send out mailer alert to Daniel and Dave in production. I'll add them to the seed file so that they're available for local test too

app/views/user_mentee_application_alert_mailer/notify_for_application_submission.html.erb
```diff
@@ -0,0 +1 @@
+<p>Application: <%= link_to "Here", user_mentee_application_url(@user_mentee_application) %>.</p>
```
GALTdea, line 1: this mail is meant for email alert for admins. I didn't think it needs any more messaging than this. but if we do I'll open another pr with it

toyhammered (approved):

Issue #227, opened 2023-08-11
[UserMenteeApplication] Accepting an applicant to join the Agency of Learning


toyhammered:
When the full flow is done and we have "accepted" someone to join the Agency of Learning 2 things should happen:
1. An email should be sent to the user letting them know they have been accepted
2. Their role should be updated to member automatically

Unsure how we want to handle rejections. 

TBD: figure out the copy of the email.

toyhammered:
@afogel This was already partially covered in https://github.com/agency-of-learning/PairApp/issues/294. I like the idea of emailing Dave first to confirm and then we can update them to the correct role. Perfectly fine if we want to manually do this via console for now and just close this ticket.

PR #305, opened 2023-09-13
294 hook up service to forms

### LINKED ISSUE #294: [UserMenteeApplication] Build real state flow into mentee application statuses
The current state flow for applications is a placeholder. We need to update this to real states the users want and the business rules for how to move between them. The full flow can be seen [here on notion](https://www.notion.so/toyhammered/Application-Submission-Design-Doc-7a45fd8e25ce4f448e9b47bc00a142b5?pvs=4#aea7590985294f589a22e6a7b70f800c)

afogel:
## What's the change?
Hooks up the new state machine managing transitions through the application process to the existing views, removes unnecessary code that was handling that before, and refines the API for the service object some.

Modal with instructions for taking an action makes more clear what should happen for each stage.

N/A

### Progressing through flow on cohort index page

[Edurrhaphy_Recording_2023_09_13-17_41.webm](https://github.com/agency-of-learning/PairApp/assets/2447409/7f41cf3b-a42e-43d2-bb7a-be11a57af237)

### Progressing through flow on application#show page

[Edurrhaphy_Recording_2023_09_13-17_42.webm](https://github.com/agency-of-learning/PairApp/assets/2447409/07b8493b-7b3f-495c-bf65-8ed61fd322b9)

Confirmation that mailers are getting sent, even when enqueued:
Screen Shot 2023-09-13 at 5 42 51 PM

Closes #294 and #227 

- [X] Have you thought of misfiring code? e.g. too many loops, n+1, or how to handle nils?
- [X] Any validations to data needed?


app/forms/mentee_application_state_form.rb
```diff
@@ -0,0 +1,21 @@
+class MenteeApplicationStateForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveModel::Validations
```
afogel, line 4: I'm probably including more than I need to. Can remove to barebones, since form isn't being used in the create controller action.

app/services/mentee_application_transition_service.rb
```diff
@@ -35,34 +35,28 @@ class InvalidTransitionError < StandardError; end
     }
   }.freeze
 
-  def promote!(application:, reviewer:, note: nil)
+  def call(application:, reviewer:, action:, note: nil)
```
afogel, line 38: simpler interface.

app/forms/mentee_application_state_form.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  attribute :action, :string
```
afogel, line 6: rationale for using a form object here is that what I care about are the actions, not the different statuses. Using the form object here enables me to leverage the form helpers/make minimal changes to how the form currently functions, but enrich it with the information we need coming from the current state.

app/services/mentee_application_transition_service.rb
```diff
@@ -35,34 +35,28 @@ class InvalidTransitionError < StandardError; end
...
   end
 
-  def reject!(application:, reviewer:, note: nil)
-    status = application.current_status.to_sym
-    raise InvalidTransitionError unless STATUS_TRANSITION_MAPPING[status][:valid_transitions].include? :reject
-    promote_transition = :rejected
-    application.mentee_application_states.create!(status: promote_transition, reviewer:, note:)
-    invoke_side_effects(application, promote_transition, reviewer)
-    true
+  def valid_transitions(status:)
```
afogel, line 54: I want to define an interface to get valid transitions to enforce message passing, rather than allowing an application to reach into the internal state of the service object.

app/forms/mentee_application_state_form.rb
toyhammered: I still don't see a purpose to this at all. It just becomes less explicit imo for what we need to pass in. I would prefer we just pass in the the 2 variables (current state and mentee application state) instead.

afogel: If I use the `MenteeApplicationState` model, I can't use the form_for helper and define an `:action` on the form, which I think is much more understandable. I think it reads much more cleanly and don't understand the harm you see in this pattern. 
Can you explain what gets obscured for you and why you think this will be confusing?

toyhammered: The obscurity is that we have this new object we have to deal with rather than just an MenteeApplicationState instance. Would you be able to show side by side what this would look like w/o the form object (maybe then i'll be able to better understand why this would be valuable?) (i'm fine if you do so in a gist or something)

afogel: https://gist.github.com/afogel/01b5900dee784f71f70b4f3b6a1f75b1

In this case, I'd be fine moving to leveraging form tags since it's a simple case, though I still think using `form_with` reads more nicely. Note: We can't use `form_with` in the option without the form object because `:action` is not a defined attribute on the application state model. The gist also contains the modifications that would be necessary for the controller to make it work.

In general, however, I think this form object pattern provides strong value when the complexity of the forms grows and/or the form straddles at least two resources.

app/services/mentee_application_transition_service.rb
```diff
@@ -35,34 +35,28 @@ class InvalidTransitionError < StandardError; end
...
-  def promote!(application:, reviewer:, note: nil)
+  def call(application:, reviewer:, action:, note: nil)
     status = application.current_status.to_sym
-    raise InvalidTransitionError unless STATUS_TRANSITION_MAPPING[status][:valid_transitions].include? :promote
-    promote_transition = STATUS_TRANSITION_MAPPING[status][:promote_transition]
-    application.mentee_application_states.create!(status: promote_transition, reviewer:, note:)
-    invoke_side_effects(application, promote_transition, reviewer)
+    # enforce that the action is cast to a symbol before performing guard clause
+    action = action.to_sym
+    raise InvalidTransitionError unless STATUS_TRANSITION_MAPPING[status][:valid_transitions].include? action
```
JoshDevHub, line 42: Since you've created `.valid_transitions`, I think it'd be nice to use it with this guard clause.

JoshDevHub (commented):
Couple small observations

toyhammered (approved):
approving so you can merge once you fix form object stuff.

app/services/mentee_application_transition_service.rb
afogel: Good call out! Will update :)

PR #307, opened 2023-09-15
Default hours to 0 on application

### LINKED ISSUE #296: [UserMenteeApplication] Set default available hours to 0
When we load the application for the user, instead of it defaulting to 1, it should default to 0 so that it forces them to consciously change it.

If possible include client side validations that allow it to start at 0, otherwise the backend validation will handle it just fine.

JoshDevHub:
## What's the change?
- Application defaults to 0 hours, forcing the user to consciously change what's in the field.

Filling out an application

One thing I thought was maybe a bit confusing is if we want to allow them to submit with this field still sitting at 0? Currently it allows this, although it'd be a super easy fix for me to make the minimum submit value greater than or equal to 1 (while still having the input default to 0).

## Demo / Screenshots
Selection_293

Closes #296

### DIFF
--- a/app/models/user_mentee_application.rb
+++ b/app/models/user_mentee_application.rb
@@ -40,7 +40,7 @@ class UserMenteeApplication < ApplicationRecord
                           }, class_name: 'MenteeApplicationState', dependent: nil
   # rubocop:enable Rails/InverseOf
 
-  validates :available_hours_per_week, numericality: { greater_than: 0, less_than_or_equal_to: 60 }
+  validates :available_hours_per_week, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 60 }
 
   validates :city, :state, :country, :reason_for_applying, :learned_to_code, :project_experience,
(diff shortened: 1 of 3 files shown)


app/models/user_mentee_application.rb
```diff
@@ -40,7 +40,7 @@ class UserMenteeApplication < ApplicationRecord
                           }, class_name: 'MenteeApplicationState', dependent: nil
   # rubocop:enable Rails/InverseOf
 
-  validates :available_hours_per_week, numericality: { greater_than: 0, less_than_or_equal_to: 60 }
+  validates :available_hours_per_week, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 60 }
```
toyhammered, line 43: So the idea here was to have it still validate at `greater_than: 0` and force them to consciously put in a number without just using the default. I'm kind of not opposed though to just allowing 0 lol.

toyhammered (approved):

afogel (approved):
lgtm!

PR #323, opened 2023-09-18
only show action if application in review

### LINKED ISSUE #320: [UserMenteeApplication] Once Accepted, no additional actions needed (show page)
Image

There should be no Manage Application url because they have been accepted already.

afogel:
## What's the change?
Updates a link on the show page to only show the action in the event that the application is in review

Better polish/guardrails on application management from the show page.
Closes #320

### DIFF
--- a/app/views/user_mentee_applications/mentee_application_states/create.turbo_stream.erb
+++ b/app/views/user_mentee_applications/mentee_application_states/create.turbo_stream.erb
@@ -17,3 +17,13 @@
     %>
   <% end %>
 <% end %>
+
+<%= turbo_stream.update "#{dom_id(@user_mentee_application)}--manage" do %>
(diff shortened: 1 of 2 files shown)


app/views/user_mentee_applications/mentee_application_states/create.turbo_stream.erb
```diff
@@ -17,3 +17,13 @@
     %>
   <% end %>
 <% end %>
+
+<%= turbo_stream.update "#{dom_id(@user_mentee_application)}--manage" do %>
```
afogel, line 21: it's a little bit of duplication, but more explicit and simpler than having to leverage i18n in this case.

JoshDevHub (approved):
LGTM 🚀

PR #326, opened 2023-09-18
authorize index page access based on whether user is an applicant

### LINKED ISSUE #316: [UserMenteeApplication] Members/Admin should not be allowed to create an application
Image

As an admin I was able to login and directly go to `user_mentee_applications` and was told there is an open cohort. In addition I was able to fill out an application (which I shouldn't be).

afogel:
## What's the change?
Only authorize applicants to view UserMenteeApplications#index

That page isn't relevant to admins the information can be accessed using specific cohort show pages.

closes #316

### DIFF
--- a/app/controllers/user_mentee_applications_controller.rb
+++ b/app/controllers/user_mentee_applications_controller.rb
@@ -3,6 +3,7 @@ class UserMenteeApplicationsController < ApplicationController
   before_action :set_active_cohort, only: %i[index new create]
 
   def index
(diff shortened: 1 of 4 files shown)


JoshDevHub (commented):
I wonder if this might be a better usecase for expanding and using the [`UserOnlyPolicy`](https://github.com/agency-of-learning/PairApp/blob/main/app/policies/user_only_policy.rb) (because this check doesn't care about a record). And the call would look like:
```rb
authorize :user_only_policy, :applicant?
```

toyhammered (approved):

PR #329, opened 2023-09-19
Add moderator user role

### LINKED ISSUE #321: Add a new user role that is between member and admin
There should be a new role that allows for specific people to view/promote/reject applicants. Unsure of what the name should be.

JoshDevHub:
## What's the change?
- Adds moderator role for users
  - This role exists between members and admins
  - The difference between their permissions and regular members is that they can view applications and manage application state.
- Policies and authorization around user applications altered to incorporate new moderator role

Viewing and reviewing applications for moderator users.

Closes #321 


app/views/layouts/_navbar.html.erb
```diff
@@ -21,7 +21,7 @@
         </li>
       <% end %>
 
-      <% if user_signed_in? && (current_user.member? || current_user.admin?) %>
+      <% if user_signed_in? && (current_user.member? || current_user.admin? || current_user.moderator?) %>
```
JoshDevHub, line 24: This logic is not great, but cleaning it up will work nicely with a rename of the `member` role. Then there can be a `member?` method on the user that checks for admin, agent, or moderator.

I know I got the okay on that idea in the discord, but figured I'd put it with a separate ticket to keep control this PR's scope.

app/controllers/user_mentee_applications_controller.rb
```diff
@@ -3,7 +3,7 @@ class UserMenteeApplicationsController < ApplicationController
   before_action :set_active_cohort, only: %i[index new create]
 
   def index
-    authorize :user_only_policy, :applicant?
+    authorize :user_only, :applicant?
```
JoshDevHub, line 6: I had to change this because I made a mistake when I made the recommendation to @afogel in #326 

Policies are looked up without the `_policy` part attached. So it needs to just be `:user_only`.

app/views/layouts/_navbar.html.erb
afogel: awesome -- maybe in the separate ticket, we can also do a migration to make the enums in order of permissions, from highest to lowest (or vice versa)

afogel (approved):
nothing blocking, this looks great.

app/models/user.rb
```diff
@@ -116,6 +116,11 @@ def blog_slug
   enum role: {
     member: 0,
     admin: 1,
-    applicant: 2
+    applicant: 2,
+    moderator: 3
   }
+
+  def application_reviewer?
```
toyhammered, line 123: This should live inside the policy, not in the model.

app/views/layouts/_navbar.html.erb
toyhammered: wouldn't this be better to push into the `user_only` policy later on. As for the migration to make the enums in order I will reject that right now. Generally speaking, you don't change the numbers on those enums ever.

toyhammered (changes requested):
couple small things but looks great otherwise!

app/models/user.rb
JoshDevHub: The only problem with putting it in the `user_only` policy is that it's also needed [here](https://github.com/agency-of-learning/PairApp/pull/329/files#diff-a5be11d25ddf72ce693ca15c7de9bd8760a1923931bafdfae5e9be70d27f0978R7) inside of the `UserMenteeApplicationPolicy`.

I'd rather avoid duplicating the method definition across multiple policies. Another solution would be to delegate the call in `UserMenteeApplicationPolicy` to `UserOnlyPolicy` ie:
```rb
def show?
  application_reviewer? || matching_user?
end

private

def application_reviewer?
  # imo this is a bit verbose for what's being expressed
  # wouldn't hate cleaning it up with some inheritance or mixin strategy if this will be a frequent pattern
  Pundit.policy(user, :user_only).application_reviewer?
end
```

app/views/layouts/_navbar.html.erb
afogel: in that case, the order clause on the `has_one :current_state` on `UserMenteeApplication` should be on `created_at`, not status, right?

toyhammered: it can be on either for that one cause we explicitly changed that already.

app/models/user.rb
toyhammered: I'd move this to the ApplicationPolicy base class honestly for now. We can figure out how to clean this up more later on.

toyhammered (approved):

PR #339, opened 2023-09-21
cleanup readme

toyhammered:
I want to ideally move more of the stuff to the wiki and make our readme pretty slim (and have it pointing to parts of our wiki like we did for the setting up for specific os.


afogel:
These changes look good to me, but I would make an explicit call out to further peruse the wiki to learn more about the project and how we approach development.

I went ahead and updated the wiki to include the 12 factor approach. I don't think we need the devise invitable documentation, even in the wiki, since it's a straightforward flow that can be looked up through devise.

afogel (approved):

Issue #318, opened 2023-09-17
[UserMenteeApplicatio] Accept/Deny table list buttons don't work


toyhammered:
![Image](https://user-images.githubusercontent.com/1283983/268534633-68446202-cf6b-460f-953f-6c55042175d6.png)

Those 2 buttons on this list don't work as expected.

GALTdea:
I can work on this

PR #341, opened 2023-09-21
#318 user mentee app index table tabs

GALTdea:
## What's the change?
Implements UserMenteeApplicationCohort table filter links for “In-review”, “Accepted”, “Rejected”

Allows admin users to filter  user mentee applications table

## Demo / Screenshots

https://github.com/agency-of-learning/PairApp/assets/16809030/c52b6856-8e42-42bd-9a32-17232e0b497d

#318 
- [ ] Related to readability and consistency: Did you add comments and/or documentation? (README, Notion, etc.)


JoshDevHub (changes requested):
This is what I see when I visit a cohort:
image

This is confusing to me because although the 'In Review' tab is highlighted, an accepted application is showing up here.

I think a good way to solve this would be to make 'In Review' the default filter.

db/migrate/20230921204534_remove_status_from_user_mentee_application.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveStatusFromUserMenteeApplication < ActiveRecord::Migration[7.0]
+  def change
+    remove_column :user_mentee_applications, :status, :integer
+  end
```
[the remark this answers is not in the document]
GALTdea: Oh yes, I wasn't supposed to push this changes, i was testing something. This will be removed next push. 

I just submitted a different PR to remove this column that didn't include a migration

app/controllers/user_mentee_application_cohorts_controller.rb
```diff
@@ -8,7 +8,24 @@ def index
...
+
+  private
+
+  def filter_applications(applications)
+    if params[:filter]
+      case params[:filter]
+      when 'in_review'
+        applications.select(&:in_review?)
+      else
+        applications.select { |application| application.current_status == params[:filter] }
```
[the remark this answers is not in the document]
GALTdea: with using where, my only concern here is that in_review? and current_status are not actual columsn in the table. so I'll have to digg in a little to figure out how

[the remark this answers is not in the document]
GALTdea: ok, given the complexity of this query, I think this only possible  with raw sql. 

i'd this the current approach Is reasonable enough

db/migrate/20230921204534_remove_status_from_user_mentee_application.rb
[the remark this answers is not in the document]
toyhammered: just pull the latest master into your branch and it will fix this.

app/views/user_mentee_applications/_table_row.html.erb
```diff
@@ -9,13 +9,13 @@
...
     <%= turbo_frame_tag "#{dom_id(mentee_application)}--note" do %>
-      <%= mentee_application.current_state.note %>
+      <%= mentee_application.current_state&.note %>
     <% end %>
   <% end %>
   
   <%= render TableCellComponent.new(container_link: mentee_application) do %>
     <%= turbo_frame_tag "#{dom_id(mentee_application)}--status" do %>
-      <%= mentee_application.status.humanize %>
+      <%= mentee_application.mentee_application_states.last.status.humanize %>
```
toyhammered, line 18: isn't this just the `current_state` you want?

app/controllers/user_mentee_application_cohorts_controller.rb
```diff
@@ -8,7 +8,24 @@ def index
...
-              .includes(:user_mentee_applications)
+              .includes(user_mentee_applications: :mentee_application_states)
               .find(params[:id])
+
+    @filtered_applications = filter_applications(@cohort.user_mentee_applications)
+  end
+
+  private
+
+  def filter_applications(applications)
```
toyhammered, line 19: This is fine for now. I'd need to pull it down locally to see but I am pretty sure we can do some ActiveRecord joins on scopes to optimize this. Not worth the trouble for now though.

toyhammered (commented):
This is looking fantastic!

app/views/user_mentee_applications/_table_row.html.erb
GALTdea: yup, fixed

app/controllers/user_mentee_application_cohorts_controller.rb
GALTdea: I'd definitely like to give it a try in a future Issue.

app/controllers/user_mentee_application_cohorts_controller.rb
```diff
@@ -8,7 +8,24 @@ def index
...
+    if params[:filter]
+      case params[:filter]
+      when 'in_review'
+        applications.filter(&:in_review?)
+      else
+        applications.filter { |application| application.current_status == params[:filter] }
+      end
+    else
+      applications.filter(&:in_review?)
+    end
```
JoshDevHub, line 29: I think the logic could be simpler here. Good to notice with conditionals when conditions or body expressions are being repeated, like with `applications.filter(&:in_review?)`

Could maybe have something like:
```rb
case params[:filter]
when 'accepted', 'rejected'
  applications.filter { |application| application.current_status == params[:filter] }
else
  applications.filter(&:in_review?)
end
```

JoshDevHub (approved):
Looks good. Just one small comment. I think it's good to merge after addressing that.

app/controllers/user_mentee_application_cohorts_controller.rb
GALTdea: ah good call, that's definitely redundant logic! Thank you

PR #356, opened 2023-09-23
remove status column from user_mentee_applications table

GALTdea:
## What's the change?
This PR removes an unneeded table column “status” in the user_mentee_applications table. Not migration referenced it.


toyhammered:
you need a migration to remove this.

afogel:
Is there an associated migration with this?

toyhammered:
> Is there an associated migration with this?

just checked and no... prob got accidentally added so I musth ave missed it in the review. Not seeing it on prod either (via ssh).

toyhammered (approved):

Issue #343, opened 2023-09-22
[UserMenteeApplicationCohort] Last Action date


toyhammered:
![Image](https://github.com/agency-of-learning/PairApp/assets/1283983/060545cf-b06b-435f-b040-716e05e2b2f4)

On this screen, add another column that shows the last action date so we can best know when we last contacted them.

GALTdea:
@toyhammered, I can tackle this!

PR #364, opened 2023-09-27
#343 add lat action col to application cohort table

GALTdea:
## What's the change?
Adds 'action_created_at'column to the user mentee application table 
Allows admins to verify when was the last/current action created on applications

## Demo / Screenshots
Screen Shot 2023-09-27 at 12 10 54 PM

#343 


app/views/user_mentee_applications/_table_row.html.erb
```diff
@@ -30,4 +30,10 @@
       <% end %>
     <% end %>
   <% end %>
+
+  <%= render TableCellComponent.new(container_link: mentee_application)  do %>
+    <%= turbo_frame_tag "#{dom_id(mentee_application)}--last_action_at" do %>
+      <%= mentee_application.last_action_created_at.to_fs(:slash_format) %>
```
toyhammered, line 36: Can this include a time also?

GALTdea: sure!

toyhammered (approved):

PR #365, opened 2023-09-27
#340, remove past applications heading when no applitions are present

### LINKED ISSUE #340: [UserMenteeApplication] Empty State for Past Applications
After signing up to become a member, applicants will see this page:

Image

Because the 'Past Applications' section is empty, having that heading present feels off. If users don't have any past applications, this heading should not show on the page.

chahmedejaz:
## What's the change?

- This PR adds the conditional rendering of the _Past Applications_ heading.
- Now we will only display the above-mentioned heading only when applications are present
- Otherwise we will hide it.

- index page for the `/user_mentee_applications`

- N/A

## Demo / Screenshots
image

Resolves #340

### DIFF
--- a/app/views/user_mentee_applications/index.html.erb
+++ b/app/views/user_mentee_applications/index.html.erb
@@ -15,12 +15,14 @@
     <% end %>
 
 
-    <h2 class="text-xl font-bold mb-2">Past Applications</h2>
-    <ul>
-      <% @user_mentee_applications.each do |mentee_application| %>
-        <% next if mentee_application.active? && mentee_application.in_review? %>
-        <%= render UserMenteeApplication::ListItemComponent.new(mentee_application:) %>
-      <% end %>
-    </ul>
+    <% if @user_mentee_applications.size.positive? %>
+      <h2 class="text-xl font-bold mb-2">Past Applications</h2>
+      <ul>
+        <% @user_mentee_applications.each do |mentee_application| %>
+          <% next if mentee_application.active? && mentee_application.in_review? %>
+          <%= render UserMenteeApplication::ListItemComponent.new(mentee_application:) %>
+        <% end %>
+      </ul>
+    <% end %>
   </div>
 </div>
--- a/app/views/user_mentee_applications/index.html.erb
+++ b/app/views/user_mentee_applications/index.html.erb
@@ -15,10 +15,11 @@
     <% end %>
 
 
-    <% if @user_past_applications.present? %>
+    <% if @user_mentee_applications.present? %>
       <h2 class="text-xl font-bold mb-2">Past Applications</h2>
       <ul>
-        <% @user_past_applications.each do |mentee_application| %>
+        <% @user_mentee_applications.each do |mentee_application| %>
+          <% next if mentee_application.active? && mentee_application.in_review? %>
           <%= render UserMenteeApplication::ListItemComponent.new(mentee_application:) %>
         <% end %>
       </ul>
--- a/app/views/user_mentee_applications/index.html.erb
+++ b/app/views/user_mentee_applications/index.html.erb
@@ -15,7 +15,7 @@
     <% end %>
 
 
-    <% if @user_mentee_applications.size.positive? %>
+    <% if @user_mentee_applications.present? %>
       <h2 class="text-xl font-bold mb-2">Past Applications</h2>
       <ul>
         <% @user_mentee_applications.each do |mentee_application| %>
--- a/app/views/user_mentee_applications/index.html.erb
+++ b/app/views/user_mentee_applications/index.html.erb
@@ -15,11 +15,10 @@
     <% end %>
 
 
-    <% if @user_mentee_applications.present? %>
+    <% if @user_past_applications.present? %>
       <h2 class="text-xl font-bold mb-2">Past Applications</h2>
       <ul>
-        <% @user_mentee_applications.each do |mentee_application| %>
-          <% next if mentee_application.active? && mentee_application.in_review? %>
+        <% @user_past_applications.each do |mentee_application| %>
           <%= render UserMenteeApplication::ListItemComponent.new(mentee_application:) %>
         <% end %>
       </ul>
--- a/app/controllers/user_mentee_applications_controller.rb
+++ b/app/controllers/user_mentee_applications_controller.rb
@@ -12,6 +12,7 @@ class UserMenteeApplicationsController < ApplicationController
                                   :mentee_application_states
                                 )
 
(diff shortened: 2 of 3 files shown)


app/views/user_mentee_applications/index.html.erb
```diff
@@ -15,12 +15,14 @@
...
 
 
-    <h2 class="text-xl font-bold mb-2">Past Applications</h2>
-    <ul>
-      <% @user_mentee_applications.each do |mentee_application| %>
-        <% next if mentee_application.active? && mentee_application.in_review? %>
-        <%= render UserMenteeApplication::ListItemComponent.new(mentee_application:) %>
-      <% end %>
-    </ul>
+    <% if @user_mentee_applications.size.positive? %>
```
afogel, line 18: use `present?` instead of `size.positive?`

afogel (changes requested):
Once the syntax is cleaned up, I think this solution would make sense here, but generally, this proposed solution would not work on other tickets requesting an empty state.

app/views/user_mentee_applications/index.html.erb
chahmedejaz: @afogel - I've addressed your comment. Can you please review?

app/views/user_mentee_applications/index.html.erb
```diff
@@ -15,12 +15,14 @@
...
 
 
-    <h2 class="text-xl font-bold mb-2">Past Applications</h2>
-    <ul>
-      <% @user_mentee_applications.each do |mentee_application| %>
-        <% next if mentee_application.active? && mentee_application.in_review? %>
-        <%= render UserMenteeApplication::ListItemComponent.new(mentee_application:) %>
-      <% end %>
-    </ul>
+    <% if @user_mentee_applications.present? %>
```
afogel, line 18: this still shows up incorrectly, as `@most_recent_application` is a subset of `@user_mentee_applications`, which means that keying in on @user_mentee_applications continues to show up.
Instead of pushing more logic into the view, I would recommend updating how `@user_mentee_applications` is assigned.

afogel (changes requested):

app/views/user_mentee_applications/index.html.erb
chahmedejaz: sure, let me get back to you on this after.

chahmedejaz: Hey @afogel - Sorry I'm a little confused here :sweat_smile:. 
- I believe it's a View's logic to display the heading (_Past Applications_) conditionally. 
- How this logic can be moved to somewhere else?
- Moreover, `@most_recent_application` empty state logic is already being handled in view as well. 
- I am confused there this can go wrong. :worried: , Please guide me. Thanks

afogel: sure -- so what you had is definitely part of the solution, but the hiccup comes when you're testing one an applicant who has already submitted an application (see your seed file for an example). 
In that case, they should have a submitted application and no past applications, but the "Past Applications" header still shows up.
So, without altering the existing view logic that you've introduced here, how can we make the issue go away?

chahmedejaz: Ohh, got it..Let me fix thatl. Thanks for the guide.

chahmedejaz: @afogel - I tried to optimize the solution by adding a scope for past applications [here](https://github.com/agency-of-learning/PairApp/pull/365/commits/3a8c2ce274b2b0b92175a50aec87ec02e9976f18). Please review.

afogel (changes requested):
it's a good effort, but this is an instance where we don't need to add more code in the form of scopes, but instead can leverage features of the ruby language to solve the problem.

I would update the index action to look like this.
```ruby
#user_mentee_applications_controller.rb
def index
    authorize :user_only, :applicant?
    @most_recent_application, *@user_mentee_applications = current_user
                                .mentee_applications
                                .order_newest_first
                                .includes(
                                  :user_mentee_application_cohort,
                                  :mentee_application_states
                                )
end
```

make sure to run the linter before updating. Good luck!

chahmedejaz:
> it's a good effort, but this is an instance where we don't need to add more code in the form of scopes, but instead can leverage features of the ruby language to solve the problem.
> 
> I would update the index action to look like this.
> 
> ```ruby
> #user_mentee_applications_controller.rb
> def index
>     authorize :user_only, :applicant?
>     @most_recent_application, *@user_mentee_applications = current_user
>                                 .mentee_applications
>                                 .order_newest_first
>                                 .includes(
>                                   :user_mentee_application_cohort,
>                                   :mentee_application_states
>                                 )
> end
> ```
> 
> make sure to run the linter before updating. Good luck!

- Oh, That's nice. :heart: 
- Sorry it was out of habit thing :sweat_smile:.
- I usually prefer database query over ruby array operations because on larger number of records, array operations just slow down the flow where as DB query is faster. 
- Apparently it's unlikely that we'd be affected by array operation here, so sure, I'll update this as per the suggestion :)

chahmedejaz:
> ```ruby
> authorize :user_only, :applicant?
>     @most_recent_application, *@user_mentee_applications = current_user
>                                 .mentee_applications
>                                 .order_newest_first
>                                 .includes(
>                                   :user_mentee_application_cohort,
>                                   :mentee_application_states
>                                 )
> ```

- Just realized while implementing, this approach has the issue. 
- We are _always_ removing the first application regardless of it being `active?` && `in_review?`
- If we have only one application, then `@user_mentee_applications` will always be empty and there will be no past applications
- Apparently, Would need to add more array operations something like `select` I guess and create the array for `past_applications`
- Please let me know what you think about it

afogel:
@chahmedejaz So I'm not sure I see an issue with the solution I'm suggesting? 
`@most_recent_application` will always be the latest application and `@user_mentee_applications` contains the rest of the applications, which by definition would also be the `past_applications`, no? Following that, `@user_mentee_applications` could be blank, but when it is, that means we shouldn't need to render the "Past Applications" header, right?

All this to say, I think adding the `if @user_mentee_applications.present?` in the view and making sure the `@user_mentee_applications` always represents all applications that aren't the most recent should satisfy the bugfix.
Or is there something here I'm misunderstanding?

chahmedejaz:
> @chahmedejaz Ah, good call out. Yes, we should handle that edge case you're suggesting, too.

@afogel - I did a fix [here](https://github.com/agency-of-learning/PairApp/pull/365/commits/1050a08a97322d9597587c9a7fea67da674cc3f8). Please let me know what you think about it. Thanks. :)

afogel (approved):

PR #370, opened 2023-09-28
#332, fixes social media links

### LINKED ISSUE #332: [UserMenteeApplication] Make working social media links
Image

Both these links should be clickable. We don't need to add the icons cause i'd prefer to see the url. They should also open up in a new tab.

chahmedejaz:
### Issue
- We are using paragraph tags for the social media links, that's the reason they were not clickable
### Solution
- This PR replaces the paragraph tags with the anchor tags so that we are able to redirect to the social media links into the new tab

- User application's view page

- N/A

## Demo / Screenshots
[Screencast from 09-28-2023 04:45:28 PM.webm](https://github.com/agency-of-learning/PairApp/assets/59338032/02e67881-6d06-4063-9ba3-e2270cd694b5)

I don't know why my mouse pointer is not showing in the video :sweat_smile:

- Resolves #332


afogel (approved):
lgtm

PR #392, opened 2023-10-11
Add mentee application state for withdrawing an application

### LINKED ISSUE #390: [MenteeApplicationState] Add Withdrawn state
Add an additional state for `withdrawn`. This state should send out a different email along the lines of "feel free to reapply when you think you are ready".

JoshDevHub:
## What's the change?
- Add `withdrawn` status to `MenteeApplicationStates` model
- Allow the application state transition service to handle applications that are transitioned to "withdrawal"
- Add mailer and notification for withdrawn applications
- Add interface for withdrawing an application

Working with applications. New ability to mark an application as "withdrawn"

## Demo / Screenshots
Selection_325

Closes #390 

## Questions:

Marking this as draft and have a few of questions I'd like answered before full review / merge
1. How's the copy on the withdrawal mailer? I kept it pretty light, not sure if we want to say more or if this is fine.
2. Do we want to introduce a new tab on the index table for withdrawn applications? Or combine them with rejected maybe? The current implementation currently doesn't display them at all, which is probably not what we want.
3. Do we want to allow applicants the opportunity to directly interact with this? To me, there's maybe no reason why an applicant couldn't be able to just directly withdraw from the process without having to email the agency and ask for it. The current implementation only allows application reviewers (mods and admins) to withdraw an application.

### DIFF
--- a/app/controllers/user_mentee_application_cohorts_controller.rb
+++ b/app/controllers/user_mentee_application_cohorts_controller.rb
@@ -18,7 +18,7 @@ class UserMenteeApplicationCohortsController < ApplicationController
 
   def filter_applications(applications)
     case params[:filter]
-    when 'accepted', 'rejected', 'application_received'
+    when 'accepted', 'rejected', 'application_received', 'withdrawn'
       applications.filter { |application| application.current_status == params[:filter] }
     else
       applications.filter { |application| application.current_status == 'application_received' }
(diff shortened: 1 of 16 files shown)


dpaola2:
1. I think the copy looks fine, ship it
2. A new tab would be helpful - otherwise, it would be hidden from view, yes?
3. At first glance I think this is a good idea, but I also wonder about side effects. For example, what if an applicant withdraws after scheduling a call?

JoshDevHub:
Awesome! Thanks for the answers.

I hadn't considered the side effects regarding your third point. I figure there can be some background stuff to clean things like that up if someone withdraws, but that'd blow the scope of this PR up. I'll just stick with mods/admins being able to withdraw for now :+1:

toyhammered (approved):

Issue #382, opened 2023-10-03
[UserMenteeApplication] Application Received should have separate tab


toyhammered:
![Image](https://github.com/agency-of-learning/PairApp/assets/1283983/f2ab96aa-c95d-4939-add4-53d29a1e9a0f)

Should add another tab (after in review maybe? could also be before idk). Just want an easier way to filter those.

GALTdea:
I'll be working on this one @toyhammered

PR #328, opened 2023-09-18
Add social media validations to mentee application model for mentee application form

elhalvers:
## What's the change?
Added social media validations to user_mentee_application model for mentee application form
## What key workflows are impacted?
mentee application form

## Demo / Screenshots
https://github.com/agency-of-learning/PairApp/assets/74928397/4cbe89ab-7da2-48ef-bf0a-06481e31ee92

## Issue ticket number and link
#314 

## Checklist before requesting a review
- Added specs for social media validations and also for the 'available_hours_per_week' validation
- Did not add a spec for:

`validates :city, :state, :country, :reason_for_applying, :learned_to_code, :project_experience,
    :available_hours_per_week, presence: true`

PR #393, opened 2023-10-11
#382 applications received tab

GALTdea:
## What's the change?
Fixes bug introduced in #328 
Adds in_review case in filter_applications method


GALTdea:
quick question @toyhammered 
I am currently using the in_rerview? method for in_review? tab which return All Applications except for accepted and rejected. 
Do we also want to exclude the "application_received" from in_review filter?

toyhammered:
> quick question @toyhammered I am currently using the in_rerview? method for in_review? tab which return All Applications except for accepted and rejected. Do we also want to exclude the "application_received" from in_review filter?

yeah we can remove it from the `in_review?` method. The application_submitted should not be shown there anymore.

PR #395, opened 2023-10-12
Applications received tab fix

GALTdea:
## What's the change?
fixes cohort application actions bug

### DIFF
--- a/app/controllers/user_mentee_application_cohorts_controller.rb
+++ b/app/controllers/user_mentee_application_cohorts_controller.rb
@@ -20,6 +20,8 @@ class UserMenteeApplicationCohortsController < ApplicationController
     case params[:filter]
     when 'accepted', 'rejected', 'application_received', 'withdrawn'
       applications.filter { |application| application.current_status == params[:filter] }
+    when 'in_review'
+      applications.filter(&:in_review?)
     else
       applications.filter { |application| application.current_status == 'application_received' }
     end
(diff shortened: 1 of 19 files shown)


JoshDevHub (changes requested):
Looks good, but the change has to be made in multiple places. Not sure how to do this in vscode, but you should be able to full project search `in_review?`. *All* of these need to use your new `action_available?` method except for the calls that are related to the tab filtering on the cohort show page.

[Here's another example of a place that needs to change](https://github.com/agency-of-learning/PairApp/blob/main/app/views/user_mentee_applications/index.html.erb#L10). If we don't change this, then users who are on the `application_received` step won't see their application here, because that step is no longer part of `in_review?`. And this isn't the only place this needs to change. Definitely recommend doing like I wrote above and just project searching calls for `in_review?`.

GALTdea:
> Looks good, but the change has to be made in multiple places. Not sure how to do this in vscode, but you should be able to full project search `in_review?`. _All_ of these need to use your new `action_available?` method except for the calls that are related to the tab filtering on the cohort show page.
> 
> [Here's another example of a place that needs to change](https://github.com/agency-of-learning/PairApp/blob/main/app/views/user_mentee_applications/index.html.erb#L10). If we don't change this, then users who are on the `application_received` step won't see their application here, because that step is no longer part of `in_review?`. And this isn't the only place this needs to change. Definitely recommend doing like I wrote above and just project searching calls for `in_review?`.

Ah, i meant to run a global search and completely got distracted. Will do! thanks for the feedback! And thank you for the feedback above? Using the value_transition method was definitely   a great idea

JoshDevHub (approved):
I might have some quibbles here but I think this works properly and it's important to just get this merged. Did some manual QA and things seem to be functioning well.

PR #398, opened 2023-10-15
Order Applications by latest updated

### LINKED ISSUE #383: [UserMenteeApplication] Cohorts Application page should be ordered by UpdatedAt
Image

The applications submitted should be ordered by the last updated on the top.

JoshDevHub:
## What's the change?
- Order applications by latest updated on the cohort show page
  - Uses `touch` strategy where new application states update the application's `updated_at` column
  - `touch` strategy requires a migration to sync the currently existing applications with their current state
  - Refactors some of the turbo behavior. The goal with this was to have it so after a reviewer actions an application from the table, that application will move to the top of the list (because it's now the most recently updated).

Viewing applications as an application reviewer

Migrate down to cleanup

## Demo / Screenshots
https://github.com/agency-of-learning/PairApp/assets/88392688/7f985ae2-f923-4e78-9ca4-65a896a513fa

Closes #383 


db/migrate/20231015020025_update_mentee_application_updated_at.rb
```diff
@@ -0,0 +1,12 @@
+class UpdateMenteeApplicationUpdatedAt < ActiveRecord::Migration[7.0]
+  def up
+    UserMenteeApplication.all.each do |application|
+      new_updated_at = application.current_state.updated_at
+      application.update!(updated_at: new_updated_at)
```
JoshDevHub, line 5: This is all a bit iffy. N+1 and each update is a separate transaction, but I couldn't immediately ideate a better way to do it while not drifting out of the scope of this ticket. See more in [this thought process doc](https://www.notion.so/toyhammered/UserMenteeApplication-Ordering-b676a58205544cc68c1aa16e268fbac4?pvs=4#95c17b6c01964f04904fc8222933bc0e)

With this only being run once against a currently pretty limited dataset, I don't think it should matter much?

toyhammered: Typically, you do these inside a rake task because you don't want to slow down a deploy by having it run this type of migration (which could fail). 

For our limited usage, this will work fine though.

toyhammered (approved):

PR #403, opened 2023-10-30
Rename "members" scope to "agents"

JoshDevHub:
## What's the change?
- Rename `agents` scope to `members`
- Ensure "moderators" are part of the scope (they weren't before)

I'd originally thought about renaming the "member" role to "agent", but after some consideration, I think I like this more. The "member" role stays as is, and "agents" would be all users who are admins, moderators, or members (ie. people who have been accepted into the agency).

If this is a good route to take, a natural next step could be to create a new `UserOnly` policy method: `#agent?`. This can be used to authorize certain actions/routes that should be agent only. I didn't include all that in this PR just because I want to make sure people are on board with this name scheme before doing that work (also keeps this PR's scope small)

Not user facing

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -85,7 +85,15 @@ class User < ApplicationRecord
 
   after_create :create_profile!
 
-  scope :members, -> { where(role: %i[member admin]) }
+  enum role: {
+    member: 0,
+    admin: 1,
+    applicant: 2,
+    moderator: 3
+  }
+  AGENT_ROLES = %w[member admin moderator].freeze
+
+  scope :agents, -> { where(role: AGENT_ROLES) }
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -85,7 +85,15 @@ class User < ApplicationRecord
 
   after_create :create_profile!
 
-  scope :members, -> { where(role: %i[member admin]) }
+  enum role: {
+    member: 0,
+    admin: 1,
+    applicant: 2,
+    moderator: 3
+  }
```
GALTdea, line 93: I think i like this set up better with the 'member' role being in the hash.

GALTdea:
Overall, It looks good and All test passing locally!

toyhammered (approved):

PR #404, opened 2023-11-01
Standup rich text

GALTdea:
## What's the change?
- Adds Rich text ability to stand up form
It gives users filling out stand up updates "Rich Text" 
## Demo / Screenshot
Screenshot 2023-11-07 at 11 01 05 AM

Screenshot 2023-11-07 at 11 01 19 AM

### DIFF
--- a/app/controllers/standup_meetings/completions_controller.rb
+++ b/app/controllers/standup_meetings/completions_controller.rb
@@ -18,6 +18,6 @@ class StandupMeetings::CompletionsController < ApplicationController
 
   def standup_meeting_params
     params.require(:standup_meeting).permit(:meeting_date, :yesterday_work_description, :today_work_description,
-      :blockers_description, :status, :yesterdays_description, :todays_description, :blockers)
+      :blockers_description, :status)
   end
 end
--- a/app/controllers/standup_meetings/completions_controller.rb
+++ b/app/controllers/standup_meetings/completions_controller.rb
@@ -18,6 +18,6 @@ class StandupMeetings::CompletionsController < ApplicationController
 
   def standup_meeting_params
     params.require(:standup_meeting).permit(:meeting_date, :yesterday_work_description, :today_work_description,
-      :blockers_description, :status)
+      :blockers_description, :status, :yesterdays_description, :todays_description, :blockers)
   end
 end
--- a/lib/tasks/migrate_fields_to_rich_text.rake
+++ b/lib/tasks/migrate_fields_to_rich_text.rake
@@ -1,0 +1,25 @@
+namespace :standup_meeting do
+  desc 'Migrate standup meeting data to action text'
+
+  task migrate_data_to_action_text: :environment do
+    failed_updates = []
+
+    StandupMeeting.where(status: 'completed').find_each do |meeting|
+      meeting.update!(
+        today_work_description: meeting.read_attribute_before_type_cast(:today_work_description),
+        yesterday_work_description: meeting.read_attribute_before_type_cast(:yesterday_work_description),
+        blockers_description: meeting.read_attribute_before_type_cast(:blockers_description)
+      )
+    rescue StandardError => e
+      failed_updates << { id: meeting.id, error: e.message }
+      Rails.logger.error "Migration failed for StandupMeeting id: #{meeting.id}, error: #{e.message}"
+    end
+
+    if failed_updates.any?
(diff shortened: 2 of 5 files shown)


app/controllers/standup_meetings/completions_controller.rb
```diff
@@ -18,6 +18,6 @@ def create
 
   def standup_meeting_params
     params.require(:standup_meeting).permit(:meeting_date, :yesterday_work_description, :today_work_description,
-      :blockers_description, :status)
+      :blockers_description, :status, :yesterdays_description, :todays_description, :blockers)
```
toyhammered, line 21: how come we have both of these now? Seems like too many params duplicated?

JoshDevHub:
I think it'd be cool to see a screenshot of what it looks like when viewing someone's standup meeting content.

app/controllers/standup_meetings/completions_controller.rb
GALTdea: Good point, fixed!

GALTdea:
> I think it'd be cool to see a screenshot of what it looks like when viewing someone's standup meeting content.

Yes, good point. Just added

lib/tasks/migrate_fields_to_rich_text.rake
```diff
@@ -0,0 +1,25 @@
...
+
+  task migrate_data_to_action_text: :environment do
+    failed_updates = []
+
+    StandupMeeting.where(status: 'completed').find_each do |meeting|
+      meeting.update!(
+        today_work_description: meeting.read_attribute_before_type_cast(:today_work_description),
+        yesterday_work_description: meeting.read_attribute_before_type_cast(:yesterday_work_description),
+        blockers_description: meeting.read_attribute_before_type_cast(:blockers_description)
+      )
```
toyhammered, line 12: i'm kind of confused as to what this exactly does?

JoshDevHub: This assigns the information that's currently in the `today_work_description`, `yesterday_work_description`, and `blockers_description` columns and assigns them as new `action_rich_texts` records associated with the standup meeting.

The new `has_rich_text` associations are set with the same name as the old columns. I also didn't expect this to work, but I just ran it locally and it does lol. I guess there's something funky with the way the reads/writes work regarding rich text associations having the same name as a column.

I probably would've done something like
- Introduce rich text associations with a different name than the current columns
- Move old column's information over to rich text
- Remove old columns (this should be done anyway, albeit in a separate PR from this one of course)
- If I wanted the rich text association names to be the same as the old column names, write a new task that does `ActionText::RichText.where(record_type: "StandupMeeting")` to pull the standup rich text records and then change their `name` attributes to match the old column names. And then make that adjustment in the standup meeting model file

That's all more complicated than what Gustavo has here, but also makes more sense to me lol.

toyhammered: > This assigns the information that's currently in the `today_work_description`, `yesterday_work_description`, and `blockers_description` columns and assigns them as new `action_rich_texts` records associated with the standup meeting.
> 
> The new `has_rich_text` associations are set with the same name as the old columns. I also didn't expect this to work, but I just ran it locally and it does lol. I guess there's something funky with the way the reads/writes work regarding rich text associations having the same name as a column.
> 
> I probably would've done something like
> 
> * Introduce rich text associations with a different name than the current columns
> * Move old columns' information over to rich text
> * Remove old columns (this should be done anyway, albeit in a separate PR from this one of course)
> * If I wanted the rich text association names to be the same as the old column names, write a new task that does `ActionText::RichText.where(record_type: "StandupMeeting")` to pull the standup rich text records and then change their `name` attributes to match the old column names. And then make that adjustment in the standup meeting model file to the `has_rich_text` associations.
> 
> That's all more complicated than what Gustavo has here, but also makes more sense to me lol.

What you just explained above is exactly how I would have done it also, but if this works, thats fine by me lol

toyhammered (approved):
I'll try to merge in and deploy to prod tomorrow + run the rake task.

lib/tasks/migrate_fields_to_rich_text.rake
GALTdea: Sorry @toyhammered, didn't get a chance to address your question. But Thanks @JoshDevHub for getting to it! 
Also, forgot to submit a chained pr with the migration to remove the old fields. I'm just going to wait until this one is merged!

Issue #416, opened 2023-11-07
[UserMenteeApplication-2.0][Admin] Move the Cohorts (index/show) pages under admin namespace


toyhammered:
Right now we have both the cohorts (index/show) as regular routes. We want to move those under a new admin namespace (and make sure only admins can view those using pundit).

PR #425, opened 2023-11-09
Move the Cohorts (index/show) pages under admin namespace

Natblow:
## What's the change?
With @Caleb-Mitchell we moved the cohort index/show into an admin interface.

Routes.
## Cleanup
We had to rename all the models/views/controllers into namespace, as well as in the Specs
## Demo / Screenshots
image

#416 


app/models/admin/user_mentee_application_cohort.rb
```diff
@@ -9,7 +9,7 @@
 #  created_at        :datetime         not null
 #  updated_at        :datetime         not null
 #
-class UserMenteeApplicationCohort < ApplicationRecord
+class Admin::UserMenteeApplicationCohort < ApplicationRecord
```
toyhammered, line 12: So we shouldn't be changing the model namespace. This should remain `UserMenteeApplicationCohort` still.

toyhammered (changes requested):
This seems mostly fine minus moving the model (which kinda cascades removing some of the changes you did).

I feel like this is also partially my fault for the lack of clarity, but generally you only need to scope your controllers, not the models (multiple controllers can talk to the same model)

Caleb-Mitchell:
Great we corrected now. 

We encountered an issue on `app/views/admin/user_mentee_application_cohorts/_table_row` where we needed to update the path method. That's why we originally tried namespacing the model and it worked. 

But, actually we just need to write it like this `admin_user_mentee_application_cohort_path(cohort)` :smile: .

app/controllers/user_mentee_applications_controller.rb
```diff
@@ -67,7 +67,7 @@ def update; end
   private
 
   def set_active_cohort
-    @active_cohort = UserMenteeApplicationCohort.active
+    @active_cohort = Admin::UserMenteeApplicationCohort.active
```
toyhammered, line 70: This shouldn't be namespaced

db/seeds.rb
```diff
@@ -89,12 +89,12 @@ def generate_pair_request_data(users:, status:, pair_date: nil)
...
 
   # Current upcoming cohort
-  active_cohort = UserMenteeApplicationCohort.create!(
+  active_cohort = Admin::UserMenteeApplicationCohort.create!(
     name: 'Fall 2023',
     active_date_range: Time.zone.today..3.months.from_now
   )
 
-  original_cohort = UserMenteeApplicationCohort.create!(
+  original_cohort = Admin::UserMenteeApplicationCohort.create!(
```
toyhammered, line 97: These shouldn't either be namespaced.

lib/tasks/reapplication_notification_task.rake
```diff
@@ -1,11 +1,11 @@
 class NoActiveCohortError < StandardError; end
 
 task enqueue_reapplication_notifications: :environment do
-  active_cohort = UserMenteeApplicationCohort.active
+  active_cohort = Admin::UserMenteeApplicationCohort.active
```
toyhammered, line 4: Also these

spec/mailers/previews/mentee_application_mailer_preview.rb
```diff
@@ -29,7 +29,7 @@ def notify_applicant_of_code_challenge_approved
 
   def notify_applicant_to_reapply
     MenteeApplicationMailer.with(recipient: User.last,
-      active_cohort_name: UserMenteeApplicationCohort.active.name).notify_applicant_to_reapply
+      active_cohort_name: Admin::UserMenteeApplicationCohort.active.name).notify_applicant_to_reapply
```
toyhammered, line 32: and this

toyhammered (changes requested):
Still a few more lingering `Admin::` that shouldn't be there.

Natblow:
Thank you! Done now.

toyhammered (approved):

Issue #251, opened 2023-08-21
Members can join the same group multiple times


afogel:
**Describe the bug**
While working on #249 , I discovered that there are no validations that prohibit a member from joining a standup meeting group multiple times. I propose adding a unique index to the StandupMeetingGroupUser to enforce a single join on the db level. 

**To Reproduce**
I think the easiest way to reproduce this is in the rails console:
```ruby
sg = StandupMeetingGroup.last
user = User.last
sg.standup_meeting_groups_users.create(user:)
sg.standup_meeting_groups_users.create(user:) # this line should fail, but it doesn't
```

afogel:
@toyhammered this should be reopened until the second PR that adds the index on the db #435 is closed, no?

PR #432, opened 2023-11-13
Feature/add uniqueness validation to standup meeting group user

wwrk22:
## What's the change?
- Added model-level uniqueness constraint to prevent a user from the joining the same standup meeting group more than once.
- Another PR will implement a database-level constraint by adding a unique index.
- Included a fix for a flash message that wasn’t showing when a user joins a standup meeting group.

N/A

In the future, we should consider the different paths a request can take on the route `/standup_meeting_groups/:standup_meeting_group_id/joins`.
e.g. How do we handle a duplicate POST request to the route caused by the user refreshing the browser during the processing of the original request.

## Demo / Screenshots
Test file `spec/requests/standup_meeting_groups/join_standup_meeting_group_spec.rb` has been written to demonstrate the failure of a request to duplicate a `StandupMeetingGroupUser` record.

#251 


app/controllers/standup_meeting_groups/joins_controller.rb
```diff
@@ -7,12 +8,8 @@ def create
 
     authorize @standup_meeting_group_user, policy_class: StandupMeetingGroup::JoinPolicy
 
-    @standup_meeting_group_user.save!
```
wwrk22, line 10: According to the [API docs](https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-save-21), the only two errors that could be raised by `#save!` are:
- `ActiveRecord::RecordInvalid` due to validation failure
- `ActiveRecord::RecordNotSaved` due to an associated `before_*` callback throwing a `:abort`

If there are existing callbacks that could throw `:abort`, perhaps they could be handled at a higher level, outside the scope of an individual controller since `StandupMeetingGroups::JoinsController` by itself doesn’t seem to be capable of triggering such a scenario.

app/controllers/standup_meeting_groups/joins_controller.rb
```diff
@@ -7,12 +8,8 @@ def create
...
-
-    respond_to do |format|
-      format.turbo_stream do
-        flash.now[:notice] = "You have joined #{@standup_meeting_group.name}!"
-      end
```
wwrk22, line 15: The `respond_to` block has been removed since
1. A request received at this controller action’s route is automatically accepted as a `turbo_stream` one.
2. We’re not handling any other formats.

app/controllers/standup_meeting_groups/joins_controller.rb
```diff
@@ -7,12 +8,8 @@ def create
...
+    if @standup_meeting_group_user.save
+      flash.now[:notice] = "You have joined #{@standup_meeting_group.name}!"
     end
```
wwrk22, line 13: Because there is no redirection after the save, it seems to be the case that the user cannot attempt to join the same standup meeting group via the application UI. If there was a redirection, I’d assume that the possibility of the user refreshing the browser during the request needs to be handled. For this reason, the save failure is not explicitly handled for now.

app/models/standup_meeting_group_user.rb
```diff
@@ -18,4 +18,10 @@ class StandupMeetingGroupUser < ApplicationRecord
 
   belongs_to :standup_meeting_group
   belongs_to :user
+
+  validates :user_id,
```
toyhammered, line 22: Can you had a test for this validation in the model?

toyhammered (changes requested):
This looks great! Just one small test I'd like added.

db/migrate/20231114144505_make_indexes_unique_for_standup_meeting_group_users.rb
```diff
@@ -0,0 +1,28 @@
+class MakeIndexesUniqueForStandupMeetingGroupUsers < ActiveRecord::Migration[7.0]
+  def change
+    # Remove duplicate records.
+    unique_ids = StandupMeetingGroupUser.group(
+      :standup_meeting_group_id, :user_id
+    ).pluck('MIN(id)')
+
+    StandupMeetingGroupUser.where.not(id: unique_ids).delete_all
```
GALTdea, line 8: @toyhammered Daniel Might want this in a separate task

db/migrate/20231114144505_make_indexes_unique_for_standup_meeting_group_users.rb
```diff
@@ -0,0 +1,21 @@
+class MakeIndexesUniqueForStandupMeetingGroupUsers < ActiveRecord::Migration[7.0]
```
toyhammered, line 1: All of this needs to be put in a separate PR.
1. We add unique constraint at model level
2. We run rake task to make sure everything is uniq at db level

Then in second PR
1. we add uniq index because it is gaurenteed to work now

toyhammered (changes requested):

db/schema.rb
```diff
@@ -179,8 +179,8 @@
     t.bigint "user_id", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["standup_meeting_group_id", "user_id"], name: "index_smg_users_on_smg_id_and_user_id"
-    t.index ["user_id", "standup_meeting_group_id"], name: "index_smg_users_on_user_id_and_smg_id"
+    t.index ["standup_meeting_group_id", "user_id"], name: "index_smg_users_on_smg_id_and_user_id", unique: true
+    t.index ["user_id", "standup_meeting_group_id"], name: "index_smg_users_on_user_id_and_smg_id", unique: true
```
toyhammered, line 183: So when you remove a migration the steps need to go:
1. `rake db:rollback` (make sure it is before the migration was run
2. rerun `rake db:migrate`

This will ensure it is removed properly (because you never want to manually touch this file)

toyhammered (changes requested):

db/migrate/20231114144505_make_indexes_unique_for_standup_meeting_group_users.rb
toyhammered: sounds good! thank you

db/schema.rb
wwrk22: Ok, I ran the rollback successfully.
Schema is back to the way it was.

toyhammered (approved):
Once tests pass this looks good!

db/schema.rb
afogel: @wwrk22 I think, as long as you're addressing it in the next PR, it's fine to disable.
To disable, I'd search for the error message first to figure out the specific rubocop to disable, then add a rubocop ignore comment before for that rule and re-enable it after the offending block of code.

wwrk22: Will do. Thanks for the tip, @afogel !

GALTdea:
test

PR #441, opened 2023-11-21
Extract `new` and `create` actions from current `mentee_application_states` controller

### LINKED ISSUE #440: [UserMenteeApplication-2.0] Extract `new` and `create` actions from current `MenteeApplicationStates` controller
These actions should be scoped to admin users. This means moving them to live under the admin namespace.

Natblow:
## What's the change?
Extracting new & create actions for `mentee_application_states` to the admin namespace. 

We separated application #show between the applicant & the admins.

Application routes

Closes #440


app/views/user_mentee_applications/show.html.erb
```diff
@@ -1,7 +1,4 @@
 <div class="mx-auto p-4 my-8 max-w-prose">
-  <% if policy(:UserOnly).admin? %>
```
JoshDevHub, line 2: A lot of the red here is a nice consequence of this separation. Admin users are guaranteed to be viewing the admin stuff, and applicant viewers are guaranteed to be viewing the applicant stuff -- all enforced at the controller level. Keeps these conditional checks out of the views :muscle:

Issue #418, opened 2023-11-08
Add Commenting Functionality to Daily Standup Posts


GALTdea:
**Is your feature request related to a problem? Please describe.**
As PairApp is becoming a central part of our daily standup routine, enabling a collaborative features could be beneficial. To enhance the interactivity and support among our developer community, we propose the addition of a commenting feature for the daily standup posts.

**Describe the solution you'd like**
Users should be able to leave comments on standup updates.

**Additional context**
This issue requires a wireframe

PR #446, opened 2023-11-29
Standup comments 1

GALTdea:
## What's the change?
- Adds comment feature to Standup Meetings that allows users to leave comments in other user's standup_metings 
- Adds StandupMeetingComments model to keep track of comments made to StandupMeetings
- Adds sections controller and index to dynamically render standup_meeting_comments belonging to any given standup_comment section i.e: "yesterday_work_description," "today_work_Description," "blocker_description," etc.
- 
Allows users to interact with other users by leaving comments on their standup meeting updates
To be fully usable, I believe I should use Turbo for comments, especially to enable users to edit their comments. but this should be done in a different PR. I'll open one with those changes as soon as this one gets accepted. 

## Demo / Screenshots
Screenshot 2023-12-02 at 3 43 17 PM

https://github.com/agency-of-learning/PairApp/assets/16809030/4f72b3e3-a872-42a4-880b-b95835a5eb7d

#418 

- [ ] Related to readability and consistency: Did you add comments and/or documentation? (README, Notion, etc.)
- [ ] Is everything elevated to the highest level? (e.g., can logic be moved out of view into controllers or out of controllers into models?)
- [ ] Can (and should) any models be extracted?


app/controllers/standup_meetings/sections_controller.rb
```diff
@@ -0,0 +1,7 @@
+class StandupMeetings::SectionsController < ApplicationController
```
GALTdea, line 1: This controller (along with its view) creates a dynamic view corresponding to the appropriate question/section in the standup meeting groups. e.g:
when params: 
http://localhost:3000/standup_meetings/597/sections?section=yesterday_work_description 

Screenshot 2024-01-09 at 10 59 42 AM

app/models/standup_meeting.rb
```diff
@@ -42,4 +44,19 @@ class StandupMeeting < ApplicationRecord
...
+  def section_content(section)
+    case section
+    when 'yesterday_work_description'
+      yesterday_work_description
+    when 'today_work_description'
+      today_work_description
+    when 'blockers_description'
+      blockers_description
+    end
+  end
```
GALTdea, line 61: Used in sections index view to query the dynamically render the appropriate comments

spec/views/standup_meeting_comments/edit.html.tailwindcss_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
```
JoshDevHub, line 1: We can delete these weird `tailwindcss_spec` files. While you're at it, you can make a tech debt issue to explore what's generating these so they can be disabled.

db/migrate/20231127185636_create_standup_meeting_comments.rb
```diff
@@ -0,0 +1,10 @@
+class CreateStandupMeetingComments < ActiveRecord::Migration[7.0]
+  def change
+    create_table :standup_meeting_comments do |t|
+      t.string :name
```
JoshDevHub, line 4: I think it's a good idea to change this to something like `section_name`. It took me some poking around to figure out what `name` meant.

Also is there any reason the `name` could every be empty? My understanding is that it's very important that it exists and that it maps to the different update types (today, yesterday, blockers) -- else it can't be displayed properly.

If my understanding is correct, then at minimum I think this field needs to be `null: false`. But also we could potentially think about making it an enum field?

JoshDevHub (changes requested):
This looks and feels great to use. Huge props @GALTdea :raised_hands: 

Just have a few comments / requested changes. Let me know if you have any questions.

Also be sure to merge main and fix the conflicts.

app/controllers/standup_meetings/sections_controller.rb
```diff
@@ -0,0 +1,7 @@
...
+  def index
+    @standup_meeting = StandupMeeting.find(params[:standup_meeting_id])
+    @section = params[:section]
+    @question = @standup_meeting.send(@section)
```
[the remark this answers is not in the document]
GALTdea: Agree! I have a solution for this but I'll create a different PR for time purposes

db/migrate/20231127185636_create_standup_meeting_comments.rb
GALTdea: Thank you for the feedback. Changes made!

app/models/standup_meeting.rb
```diff
@@ -42,4 +46,23 @@ class StandupMeeting < ApplicationRecord
...
+
+  def allowed_section?(secion_name)
+    ALLOWED_SECTIONS.include?(secion_name)
+  end
```
JoshDevHub, line 67: Awesome :raised_hands:

JoshDevHub:
Let me know when you want me to review again @GALTdea

GALTdea:
> Let me know when you want me to review again @GALTdea

Thank you, @JoshDevHub! I need to fix a couple of failing tests from my latest commits, I'll let you know when I do!

PR #450, opened 2023-12-04
User Mentee Applications: add ability for admins to view all notes associated with status

### LINKED ISSUE #429: [UserMenteeApplication-2.0][Admin] Ability to view the notes of each stage for an applicant
https://www.notion.so/toyhammered/User-Mentee-Application-2-0-Application-Tracking-047499e6055f4c77998403e985a5632c?pvs=4#9d3081606e18473e84f451932adfc362

jp524:
## What's the change?
On the admin side of User Mentee Application, all reviewer notes are shown in one place using a `select` tag.
The associated Stimulus controller displays a note based on the status selected.

Updating an applicant's status.

When updating an applicant status from the `admin_user_mentee_application_path(:id)`, the `select` element is not automatically updated. Therefore a manual refresh of the page is required to display the new status and note.
I thought about adding a Turbo action to update this `select` tag but since I am using ViewComponents I don't think that is possible. 

## Demo / Screenshots
https://github.com/agency-of-learning/PairApp/assets/85654561/d364e5f9-a64d-4811-b49f-8a59e370847c

Closes #429 

### DIFF
--- a/app/components/user_mentee_application/notes_component.html.erb
+++ b/app/components/user_mentee_application/notes_component.html.erb
@@ -2,11 +2,4 @@
   <h3 class="text-lg font-semibold">Reviewer notes</h3>
 </div>
 
-<ul>
-  <% status_and_notes.map do |status, note| %>
-    <li class="mb-1">
-      <span class="font-semibold"><%= status %>: </span>
-      <%= note %>
-    </li>
-  <% end %>
-</ul>
+<%= select_tag "status", options_for_select(status_and_notes.keys, current_status), class: "select select-primary"%>
--- a/app/components/user_mentee_application/notes_component.html.erb
+++ b/app/components/user_mentee_application/notes_component.html.erb
@@ -1,0 +1,12 @@
+<div class="pb-5">
+  <h3 class="text-lg font-semibold">Reviewer notes</h3>
+</div>
+
+<ul>
+  <% status_and_notes.map do |status, note| %>
+    <li class="mb-1">
+      <span class="font-semibold"><%= status %>: </span>
+      <%= note %>
+    </li>
+  <% end %>
+</ul>
--- a/app/components/user_mentee_application/notes_component.html.erb
+++ b/app/components/user_mentee_application/notes_component.html.erb
@@ -1,5 +1,20 @@
-<div class="pb-5">
-  <h3 class="text-lg font-semibold">Reviewer notes</h3>
-</div>
 
-<%= select_tag "status", options_for_select(status_and_notes.keys, current_status), class: "select select-primary"%>
+<div class="text-md text-base-content/70 mb-3">Reviewer Notes</div>
+
+<div data-controller="select" data-select-notes-value="<%= status_and_notes.to_json %>">
+  <div>
+    <span class="mr-2">Status</span>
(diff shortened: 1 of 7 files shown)


JoshDevHub:
This is awesome! Couple comments:

1. I thiiink having this work with turbo would be as simple as re-rendering the component in the existing `create` turbo stream. [Here's the existing code to update the note section](https://github.com/agency-of-learning/PairApp/blob/main/app/views/admin/user_mentee_applications/mentee_application_states/create.turbo_stream.erb#L12-L14). Instead of:
    ```erb
      <%= @user_mentee_application.current_state.note %>
    ```
    You can render the component there. And then where you're rendering the component on the show page, make sure the `span` has an id like `"#{dom_id(@user_mentee_application)}--note"` -- this way when a new status is created, it'll know to rerender your view component in that note section. Let me know if this doesn't work or you have any questions about it.

2. This is a bit more of a theoretical design thing. You've modeled this as a select input, but I think you could also  model it as a dropdown, where each "item" is a link to say, `NotesController#show` (which you'd have to make). And this is wired up using turbo frames such that when the admin clicks a link in the dropdown, the `notes` frame would get replaced with the content for the selected item. Hopefully that makes sense :sweat_smile: 
Anyway, I'm not saying for sure this would be a "better" way of doing it, and I'd be interested in knowing others' thoughts. But I also know a lot of people view using select inputs outside of a form context for this type of usecase as a codesmell. Would be interested in @afogel and @toyhammered thoughts on this.

afogel:
@jp524 Looks really great! I think I agree w/ @JoshDevHub's assessment on using the select; it does strike me as though the affordance of that may not be as ideal, both from a UI design standpoint as well as a UX/code perspective. Like Josh mentioned, I typically think of selects in terms of forms. That having been said, I think the way you're using them isn't out of bounds.

One thing I liked about your choice to use the select element over the [accordion](https://daisyui.com/components/accordion/#using-accordion-and-join-together) is that it provides a much more streamlined/space efficient experience compared to the accordion, especially as we get further in the process. My initial instinct is that a [tabbed component](https://daisyui.com/components/tab/#radio-tab-lifted--tab-content) might be a better experience, but it still suffers from some of the issues of space efficiency and finding what's relevant as the process advances onward.

As you mentioned, it isn't ideal to have to refresh the page to see the new changes reflected in the page. One alternative interface might be to do [something like this](https://balsamiq.cloud/se3m7v2/puupn9o/rFD30#):
Screenshot 2023-12-04 at 9 26 09 PM

This is kind of like a vertical tab situation and I think would dovetail well with what Josh suggested -- when you create the note and advance the applicant, you can use a turbo frame to prepend the latest status to the box on the left hand side and replace the main content area with a turbo frame containing notes on that page. In this conceptualization, each link on the left hand side representing a previously completed stage in the application process that would fetch the associated notes and replace the main content area.

What do you think?

jp524:
Thank you @JoshDevHub and @afogel for your insights! I liked seeing your approaches to the problem and I can see now that it can definitely be solved without JavaScript.

I ended up keeping the `select` tag solution for now, and simply re-renderedthe NotesComponent through a Turbo Stream.

JoshDevHub (approved):
LGTM :rocket:

Issue #212, opened 2023-08-08
[Standup Updates] Add Emoji Reactions to Standup Posts


dpaola2:
We'd like to introduce emoji reactions for each daily standup post in PairApp to allow users to interact with updates more easily and expressively.
Emoji reactions will enable users to quickly show support, express concerns, or convey their feelings about a standup update without the need for typing out full comments.

**Describe the solution you'd like**
- Emoji reactions (👍 , 🎉 , 🤔 , 👁️ , etc.)

**Additional context**
This issue requires a wireframe and design doc
[Wireframe](https://balsamiq.cloud/se3m7v2/puupn9o/r32E2)

afogel:
This seems like a great [EPIC]!

PR #451, opened 2023-12-12
Feature/add rich text reaction model

wwrk22:
## What's the change?
Added a model that stores emojis for reacting to standup meeting updates. The backend implementation was done on this branch.

Users can create `RichTextReaction` records via the console to emulate reacting to standup meeting updates.

The frontend implementation that includes the UI to allow users to react is still required.

## Demo / Screenshots
Create reactons via the Rails console.
```
# Assume current_user is the signed-in User, and rich_text_id is the id of an ActionText::RichText record.
rich_text_id = StandupMeeting.where(user: current_user).yesterday_work_description.id

# Refer to RichTextReaction.emojis for emojis that can be used in creating reactions.
current_user.rich_text_reactions.create(rich_text_id:, emoji_caption: ‘thumbs_up')
```

#212 
This PR does not completely resolve the issue.


app/models/rich_text_reaction.rb
```diff
@@ -0,0 +1,53 @@
...
+class RichTextReaction < ApplicationRecord
+  # Permissible set of emojis for standup updates and comments.
+  EMOJIS = {
+    👍: 'thumbs-up',
+    🤔: 'thinking',
+    🎉: 'hooray',
+    🤷: 'shrug',
+    👎: 'thumbs-down',
+    👀: 'eyes'
+  }.freeze
```
wwrk22, line 31: The emoji and the caption will be flipped. After testing the code in the console, I’ve come to the realization is that a the caption being the key makes it easier for development.
i.e. It’s easier and more readable to type `EMOJIS[:thumbs_up]` then it is to type `EMOJIS[:👍]`.

app/views/rich_text_reactions/create.turbo_stream.erb
wwrk22, line 1: I didn’t know how to allow the controller action to work without having this file in place.
Would appreciate anyone’s advice for better dealing with this.

app/controllers/rich_text_reactions_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def build_rich_text_reaction
+    @rich_text_reaction = current_user.rich_text_reactions.build(
+      rich_text_reaction_params
+    )
+  end
+
+  def set_rich_text_reaction
+    @rich_text_reaction = RichTextReaction.find(params[:id])
+  rescue ActiveRecord::RecordNotFound
```
jp524, line 30: Is there a specific reason why we are rescuing the error here? Line 17 will also raise an error if the `@rich_text_reaction` can't be destroyed.

app/models/rich_text_reaction.rb
```diff
@@ -0,0 +1,54 @@
...
+      message: 'must be present in permissible set'
+    }
+
+  # Return the emoji icon for the caption.
+  # e.g. '👍' for 'thumbs_up'
+  def emoji
+    EMOJI_DICT[emoji_caption]
+  end
+
+  class << self
```
jp524, line 48: Personally I find the syntax `def self.emoji_captions` easier to understand.

jp524:
In the Demo / Screenshots section of the PR, last line should read `current_user.rich_text_reactions.create(rich_text_id:, emoji_caption: :thumbs_up)` since `emoji` was changed to `emoji_caption` in the schema.

spec/requests/rich_text_reactions_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+RSpec.describe 'RichTextReaction requests' do
+  let!(:user) { create(:user) }
+
+  before do
+    sign_in user
+  end
+
+  describe 'POST /rich_text_reactions' do
+    it 'creates a rich text reaction' do
+      expect {
```
jp524, line 12: For readability I would write the two `expect` lines at the end of the `it 'creates a rich text reaction' do` block.

spec/requests/rich_text_reactions_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      expect(flash.now[:alert]).to be_present
+    end
+  end
+
+  describe 'DELETE /rich_text_reactions/:id' do
+    it 'deletes a rich text reaction' do
+      id = create(:rich_text_reaction).id
+
+      expect {
```
jp524, line 43: Same comment as above regarding the location of the expectation.

app/views/rich_text_reactions/create.turbo_stream.erb
jp524: The file is required since the controller renders the associated `create` view on line 8. I think it's okay to leave it like this for now. Once the front end portion is in place, you can decide where you want the controller to render or redirect.

wwrk22:
> In the Demo / Screenshots section of the PR, last line should read `current_user.rich_text_reactions.create(rich_text_id:, emoji_caption: :thumbs_up)` since `emoji` was changed to `emoji_caption` in the schema.

Fixed

app/controllers/rich_text_reactions_controller.rb
wwrk22: No specific reason. I just did it out of habit of catching every error I am ever aware of when I used to work with C++ lol.
For `#destroy` on line 17, I have two questions.
1. Are you saying that `ActiveRecord::RecordNotFound` could be raised if I try to destroy a non-existing record? If so, should I also add a rescue there? I’m aware that there are clean ways to do this such as the `#rescue_from` so that I don’t repeat the rescue call.
2. The [API](/Users/wonrhim/.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/activerecord-7.0.8/lib/active_record/transactions.rb) says (let me know if this is the wrong method) `#destroy` returns false when a `before_destroy` callback throws an `:abort`. Should this be handled as well?

app/models/rich_text_reaction.rb
wwrk22: Here are two reasons why I’ve learned to prefer this style:
1. I like how `class << self` allows me to physically organize my class methods, especially with private/protected methods.
2. I like looking at class methods like I would another class definition since class methods are merely instance methods of the singleton class object for the class we’re working inside.

Having said that, I prefer consistency among team members, and would be glad to change this.
Just let me know. 👍

spec/requests/rich_text_reactions_spec.rb
wwrk22: I think I know what you’re suggesting but I want to be clear.
Are you suggesting I take out the code for `post` and run that first, and then `expect` the count to change?
Something like this?
```
count = RichTextReaction.count
post rich_text_reaction_url, …
expect(RichTextReaction.count).to eq(count + 1)
```

spec/requests/rich_text_reactions_spec.rb
wwrk22: Same question as [this one](https://github.com/agency-of-learning/PairApp/pull/451#discussion_r1435717429).

spec/requests/rich_text_reactions_spec.rb
jp524: Exactly :)

app/models/rich_text_reaction.rb
jp524: All good as is then. Thanks for the clarification!

app/controllers/rich_text_reactions_controller.rb
jp524: 1. If you try to delete a record that does not exist, line 17 will actually raise a `NoMethodError` since `@rich_text_reaction` will be nil. So you would need to add a `rescue` block in the `destroy` method too, unless you integrate the `set_rich_text_reaction` method into `destroy`. I believe this would be the way to go since `set_rich_text_reaction` is only called by `destroy` anyway.
2. As discussed - we will review this later if it becomes an issue

spec/factories/rich_text_reactions.rb
```diff
@@ -0,0 +1,28 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_...  (rich_text_id => action_text_rich_texts.id)
+#  fk_rails_...  (user_id => users.id)
+#
+FactoryBot.define do
+  factory :rich_text_reaction do
+    user
+    rich_text { create(:standup_meeting).yesterday_work_description }
```
JoshDevHub, line 25: Reminder to fix this

JoshDevHub (changes requested):
Just one reminder of something to fix in here. The other stuff looks really good though.

spec/factories/rich_text_reactions.rb
```diff
@@ -0,0 +1,35 @@
...
+#  fk_rails_...  (rich_text_id => action_text_rich_texts.id)
+#  fk_rails_...  (user_id => users.id)
+#
+FactoryBot.define do
+  factory :rich_text_reaction do
+    # +rich_text_owner+ can be given by the user of this factory, and it may
+    # be any persisted ActiveRecord model record. By default, a User record is
+    # used.
+    transient do
+      rich_text_owner { user }
```
wwrk22, line 28: @JoshDevHub 
Some ActiveRecord model object has to be given for creating ActionText::RichText.
I decided the User model is probably the safest to use since it most likely won’t be going away or changed to a different name.

spec/factories_spec.rb
```diff
@@ -2,6 +2,8 @@
 
 RSpec.describe 'FactoryBot' do
   FactoryBot.factories.each do |factory|
+    next if factory.name == :action_text_rich_text
```
wwrk22, line 5: @JoshDevHub 
I think unless we can somehow have a fake ActiveRecord model with a fake database table, it’s best to skip linting the ActionText::RichText factory.

JoshDevHub (approved):
Alright cool. Looks like you have to fix some merge conflicts and it'll be good to go

wwrk22:
Per [the comment above](https://github.com/agency-of-learning/PairApp/pull/451#pullrequestreview-1811631101):
Merge conflict has been resolved.

PR #461, opened 2024-01-10
Fix/namespace inconsistency

wwrk22:
## What's the change?
Pluralize namespaces for consistency and to avoid flooding model class namespace

N/A

N/A

## Demo / Screenshots
N/A

N/A


app/notifications/mentee_application/acceptance_notification.rb
wwrk22, line 1: @JoshDevHub 
All `_notification.rb` files were moved to `app/notifications/user_mentee_applications` for consistency since other parts of the project prefix the namespace with `user_` and the logic of these notification classes seem to be within contexts that include the `UserMenteeApplication` model.

app/policies/pair_requests/acceptance_policy.rb
wwrk22, line 1: @JoshDevHub 
I left `app/policies/*_policy.rb` files as they are because the Pundit gem doesn’t support pluralized naming.
e.g. If I change `PairRequestPolicy` to `PairRequestsPolicy` and change the call `policy_scope(PairRequest)` to `policy_scope(PairRequests)`, then it will not work.

app/workers/pair_requests/auto_expire_worker.rb
```diff
@@ -1,4 +1,4 @@
-class PairRequest
+module PairRequests
```
JoshDevHub, line 1: We have to take extra precautions when editing a worker's name. This is because when a worker is enqueued, the worker's name is one of the things that's saved into queue. When the worker is dequeued and processed, its name is used to look up the code that should be run.

This causes an issue though if we change the name while workers are in the queue. To safely rename, we have to temporarily alias the class until we're sure no workers with the old  names are present in the queue.

Can [see here](https://github.com/sidekiq/sidekiq/wiki/FAQ#how-do-i-safely-rename-a-worker) in the sidekiq wiki for how this is done. You should do this for each of the changing worker names. Then we can come back later with a new PR in ~1 week that cleans up the old names completely.

JoshDevHub (changes requested):
Looks pretty good! Thanks for doing this.

Just have one change I'm going to recommend regarding the Worker names. Let me know if you have any questions.

app/workers/pair_requests/auto_expire_worker.rb
wwrk22: Will do

wwrk22: I’ve set the aliases and a reminder for myself to remove them after a week.
Would I be able to check the queue myself?

wwrk22:
@JoshDevHub 
By the way, I’m adding tests for the classes that codecov is complaining about.
I’d like to take the time to do it because I’m learning a lot about some of the gems like Pundit and Noticed. It also looks like some of the tests might be outdated.

codecov.yml
```diff
@@ -1,2 +1,3 @@
 ignore:
   - "app/controllers"
+  - "config/initializers/scheduler.rb"
```
wwrk22, line 3: @JoshDevHub 
codecov started complaining after I appended the “s” after the namespaces in the scheduler
Is the scheduler something we need to test?

app/workers/pair_requests/auto_expire_worker.rb
JoshDevHub: I can't even view it unfortunately :(

The production queue is only viewable for admin users.

codecov.yml
JoshDevHub: Nah we shouldn't have to test anything happening in the scheduler.

spec/policies/standup_meeting_groups/join_policy_spec.rb
JoshDevHub, line 1: Thanks for writing specs on these untested policies :raised_hands:

